PR #51, opened 2018-12-28
[ignitions] set realtime scheduler using Systemd's policy instead of chrony.conf

zoetrope:
If sched_priority is specified in chrony.conf, the following log is output.
```
chrony[7]: 2018-12-27T23:00:12Z sched_setscheduler() failed
```


ymmt2005:
LGTM

PR #77, opened 2019-01-22
Use proxy for (and only for) exterior communication.

morimoto-cybozu:


pkg/neco/cmd/sabakan_upload.go
```diff
@@ -27,6 +28,16 @@ func sabakanUpload(ctx context.Context, st storage.Storage) error {
 	}
 	localClient := ext.LocalHTTPClient()
 
+	// NOTE: hack to set http proxy to container/image
```
ymmt2005: hack for github.com/containers/image to set HTTP proxy

ymmt2005:
LGTM

PR #142, opened 2019-03-03
[ignitions] add config for MetalLB to bird.conf

ymmt2005:
MetalLB runs a BGP speaker to advertise virtual IP addresses
on each Kubernetes node.  Since each node already runs bird as
BGP speaker, MetalLB need to communicate with bird instead of
ToR switches.

In Neco, ASN for MetalLB is `64698` and bound to 127.0.0.1.
BIRD advertises routes from MetalLB as its own routes by
using `next hop self` directive.

In addition, MetalLB routes should not be imported to Node
kernel because packets to the virtual IP addresses are handled
by kube-proxy (IPVS).  To distinguish MetalLB routes, they are
marked with BGP community `(64698,64699)`.

### DIFF
--- a/ignitions/common/files/etc/bird/bird.conf
+++ b/ignitions/common/files/etc/bird/bird.conf
@@ -18,6 +18,7 @@ protocol kernel {
         export filter {
             if source = RTS_DEVICE then reject;
             if proto = "coil" then reject;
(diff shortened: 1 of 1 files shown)


mitsutaka:
Ref: https://github.com/cybozu-go/neco/compare/metallb-poc

PR #174, opened 2019-03-21
Fix API group of PodSecurityPolicy

ymmt2005:
This can be merged only after CKE is released with https://github.com/cybozu-go/cke/pull/154.


zoetrope:
PodSecurityPolicy was moved to policy/v1beta1 from extensions/v1beta1

PR #196, opened 2019-04-03
Use different context when updater sends slack notification

mitsutaka:
neco-updater does not send a notification sometimes when
context is canceled due to updating or restarting.

warning: "failed to notify" error="Post https://hooks.slack.com/services/foo/bar context canceled"

Also omit "ts" field from the slack attachment JSON for removing default
time.Time in the notification message.


binoue:
LGTM

PR #260, opened 2019-04-23
add kubeapi options to run k8s-csi-lvm

satoru-takeuchi:


satoru-takeuchi:
We canceled to evaluate k8s-csi-lvm because we found that this tool doesn't satisfy our requirements as a result of reading source code.

PR #263, opened 2019-04-24
Use nilerr and restrictpkg

dulltz:


dulltz:
Waiting for the release including this fix https://github.com/golang/go/issues/26849#issuecomment-485875819

ymmt2005:
Maybe with Go 1.13?

dulltz:
Close ref: https://github.com/cybozu-go/neco/pull/818

PR #274, opened 2019-05-13
[ignition] enable lvm2-monitor.service

ymmt2005:
This service monitors LVM2 snapshots and automatically
extends COW blocks when it runs low.

### DIFF
--- a/ignitions/common/files/opt/sbin/setup-var
+++ b/ignitions/common/files/opt/sbin/setup-var
@@ -25,6 +25,10 @@ lvsize() {
 }
 
 prepare_lv() {
+    # enable COW volume auto extension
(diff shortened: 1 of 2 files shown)


morimoto-cybozu:
`snapshot_autoextend_threshold` in lvm.conf is 100 in default, which disables automatic extensions.
Can we have lvm.conf or command-line configuration to set `snapshot_autoextend_threshold`?
https://dustymabe.com/2012/03/04/automatically-extend-lvm-snapshots/
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html-single/logical_volume_manager_administration/index

ymmt2005:
@morimoto-cybozu 
Thank you.  Pushed a commit to address the problem.
PTAL.

PR #335, opened 2019-06-27
Fix the problem that serf cluster is divided

dulltz:
### Reproduction Steps

On dctest environment,

1. Stop all worker nodes
1. Disable bird.service on boot-0
1. Stop boot-0
1. Stop serf.service on boot-1 and boot-2 to delete boot-0 from serf members
1. Reboot boot-0
1. Start bird.service on boot-0

After this procedure, the serf cluster is divided boot-0 and others.
If you restart serf.service on boot-0, the cluster will recover.


dulltz:
### Reproduce procedure on dctest

1. Stop all worker nodes
1. Disable bird.service on boot-0
1. Stop boot-0
1. Stop serf.service on boot-1 and boot-2 to delete boot-0 from serf members
1. Reboot boot-0
1. Start bird.service on boot-0

After this procedure, the serf cluster is divided boot-0 and others.
If you restart serf.service, the cluster will recover.

PR #336, opened 2019-06-27
[ignition] Use CoreOS 2079.4.0 for workaround

kfyharukz:


generator/generate.go
```diff
@@ -220,6 +220,10 @@ func getLatestCoreOS(ctx context.Context) (*neco.CoreOSImage, error) {
 		versions = append(versions, v)
 	}
 	sort.Sort(sort.Reverse(version.Collection(versions)))
+	if versions[0].Original() == "2135.4.0" {
+		log.Info("do not use CoreOSImage 2135.4.0 because this version's rkt is not worked", nil)
```
ymmt2005:
```suggestion
		log.Info("do not use CoreOSImage 2135.4.0 because this version's rkt does not work", nil)
```

PR #350, opened 2019-07-05
Set Restart=on-failure and StartLimitIntervalSec=600s

mitsutaka:
It prevents many restarts and creating many exited rkt containers.


progs/cke/templates.go
```diff
@@ -15,8 +15,9 @@ ConditionPathExists={{ .KeyFile }}
 Slice=machine.slice
 Type=simple
 KillMode=mixed
-Restart=always
+Restart=on-failure
 RestartSec=10s
+StartLimitInterval=10m
```
dulltz: https://github.com/systemd/systemd/blob/v241/NEWS#L3443
Move this to [Unit]

progs/cke/templates.go
```diff
@@ -10,14 +10,14 @@ After=network-online.target
 ConditionPathExists={{ .ConfFile }}
 ConditionPathExists={{ .CertFile }}
 ConditionPathExists={{ .KeyFile }}
+StartLimitInterval=10m
```
dulltz: Sorry, it seems that `StartLimitIntervalSec` more appropriate.
https://github.com/systemd/systemd/pull/3148/commits/f0367da7d1a61ad698a55d17b5c28ddce0dc265a

PR #373, opened 2019-07-12
Rename name label

dulltz:


dulltz:
Remove `k8s-app: coil-node` and `k8s-app: coil-controller` labels because they are not referred from neco-apps.
Remain `k8s-app: squid` and `k8s-app: unbound` labels because they are referred from network policies in neco-apps.

PR #393, opened 2019-07-24
[dctest] fast bootstrap

mitsutaka:
This PR has tweaks as follows:

- Analyze some running time with `time` command.
- [dctest] Split function suite to bootstrap and function to analyze bootstrap time.
- [ignition] Download cke images in parallel.
- [Makefile] `dpkg-build` without compression for testing.
- [CI] Move all deleting instance step to CircleCI for easier instance management.
- [gcp] Fix the wrong argument for attaching home disk to GCE instance.
- [Makefile] Run Makefile.tools with MAKEFLAGS -j$(nproc) to download assets for debian package in parallel.
- [dctest] Move the rebooting boot server test to function suite to reduce testing time of bootstrap suite.


.circleci/config.yml
```diff
@@ -103,7 +103,16 @@ jobs:
...
+          suite: ./bootstrap
+      - run:
+          name: run dctest functions
+          command: ./bin/run-dctest-suite.sh functions
+          no_output_timeout: 20m
+      - run:
+          name: delete instance
+          command: |
+            . ./bin/env
+            $GCLOUD compute instances delete ${INSTANCE_NAME} --zone ${ZONE}
```
satoru-takeuchi: `when: always` is necessary since this step is skipped if one of the above steps fails.
Before this commit, there is no such leaking since the instance is deleted in the
cleanup routine.

This comment is applied to all the remaining "delete instance" targets.

mitsutaka: If I set `when: always`, test instance is always deleted. So we won't be able to debug.

`delete instance` step should not be run when the previous step is failed for debugging. `run-dctest.sh` skipped deletion instance as follows:

```
delete_instance() {	
  if [ $RET -ne 0 ]; then	
    # do not delete GCP instance upon test failure to help debugging.
    return	
  fi	
  $GCLOUD compute instances delete ${INSTANCE_NAME} --zone ${ZONE} || true	
}
```

satoru-takeuchi: Got it, thanks.

PR #444, opened 2019-08-09
Add a new command generate-sabakan-machines

mitsutaka:
Registration of number of machines is expects huge when we deploy
a new datacenter or a new 3-rack set.

This command supports to generate the machines.json for `sabakan machines create -f`

### DIFF
--- a/pkg/generate-sabakan-machines/main.go
+++ b/pkg/generate-sabakan-machines/main.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"strconv"
+	"time"
 
 	sabakan "github.com/cybozu-go/sabakan/v2"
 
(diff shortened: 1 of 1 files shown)


pkg/generate-sabakan-machines/main.go
```diff
@@ -0,0 +1,121 @@
...
+				"machine-type": machineType,
+				"datacenter":   datacenter,
+				"product":      product,
+			},
+			Rack: uint(rackInt),
+			Role: role,
+			BMC: sabakan.MachineBMC{
+				Type: "IPMI-2.0",
+			},
+		}
```
dulltz: `retire-date` is not automatically written. Does it mean that it will be written manually later?
https://github.com/cybozu-go/sabakan/blob/master/docs/machine.md#machinespec-struct

mitsutaka: TODO: `retire-date` might be included in CSV.

mitsutaka: Generate `retire-date` from support date + 5 years.

PR #464, opened 2019-08-16
dctest: change intance type and worker memory to stabilize test

satoru-takeuchi:
There are the following two instability problem about dctest.

* dctest-function fails at about 50%.
* The test of ECK sometimes fails.

To resolve these issues, upgrade to the memory-rich instance type and give more
memory for each CSes and SSes.

I confirmed that 5 / 5 dctest-function succeeded with this fixes.


ymmt2005:
Addressed by myself.

PR #472, opened 2019-08-19
Refactor CI

masa213f:


.circleci/config.yml
```diff
@@ -301,21 +317,15 @@ workflows:
       - deploy_github:
           requires:
             - build_release
-          filters:
```
ymmt2005: Why did you remove this?

.circleci/config.yml
```diff
@@ -157,6 +163,7 @@ jobs:
           command: |
             git config --global user.email "neco@cybozu.com"
             git config --global user.name "cybozu-neco"
+      # Check difference between [`release` branch] and [`master` branch + `artifacts_release.go`]
       - run: git checkout release
       - run: git merge --no-commit origin/${CIRCLE_BRANCH}
```
ymmt2005: Fix this as well.

.circleci/config.yml
masa213f: Because `filters` is set in the dependent `build_release` job.
Should I specify it explicitly?

.circleci/config.yml
masa213f: Thank you! d1ce99a

.circleci/config.yml
ymmt2005: I suppose so.

PR #500, opened 2019-08-30
sabakan-state-setter: fix health condition for r640-boot-2

ymmt2005:


etc/sabakan-state-setter.yml
```diff
@@ -3,29 +3,6 @@ machine-types:
     grace-period: 5s
     metrics:
       - name: hw_processor_status_health
```
morimoto-cybozu: Can we remove this as well?
I suppose that this metrics rule is matched to "empty" processor status list and is evaluated to true.

ymmt2005: Ah, how about this?

```suggestion
      - name: processor_status_health
```

c.f. https://github.com/cybozu-go/setup-hw/blob/master/redfish/rules/qemu.yml#L8

ymmt2005: I think `monitor-hw` exports this metric for Qemu according to these lines:
- https://github.com/cybozu-go/setup-hw/blob/2d40467d1bed7a58e8b7a48bb766e89c5a95ff6c/redfish/mock.go#L17-L32
- https://github.com/cybozu-go/setup-hw/blob/2d40467d1bed7a58e8b7a48bb766e89c5a95ff6c/redfish/rules/qemu.yml#L5-L9

morimoto-cybozu:
Metrics names are prefixed with `hw_`, which is the 1st argument of `prometheus.BuildFQName()` in redfish/rule.go.

What I mean is that the Redfish collector for QEMU will produce no metrics because there is no Redfish data,
and that the metrics rules in neco/etc/sabakan-state-setter.yml have no meaning.
But this is my mistake.
The collector will use defaultDummyData in redfish/mock.go.

So the metrics rules are effective, and there is no need to change.

etc/sabakan-state-setter.yml
morimoto-cybozu: Metrics names are prefixed with `hw_`, which is the 1st argument of `prometheus.BuildFQName()` in redfish/rule.go.

What I mean is that the Redfish collector for QEMU will produce no metrics because there is no Redfish data,
and that the metrics rules in neco/etc/sabakan-state-setter.yml have no meaning.
But this is my mistake.
The collector will use defaultDummyData in redfish/mock.go.

So the metrics rules are effective, and there is no need to change.

PR #536, opened 2019-09-10
[ignitions] Add systemd units for garbage collection of containers and container images

dulltz:


ignitions/common/systemd/docker-prune.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=run docker prune periodically
+PartOf=docker-prune.service
```
ymmt2005: What the intension of this `PartOf`?

I cannot find any existing example:
```console
/lib/systemd/system$ ls *.timer
apt-daily.timer          fstrim.timer              motd-news.timer               ureadahead-stop.timer
apt-daily-upgrade.timer  mdadm-last-resort@.timer  systemd-tmpfiles-clean.timer

/lib/systemd/system$ grep PartOf *.timer
```

ignitions/common/systemd/rkt-gc.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=run rkt gc periodically
+PartOf=rkt-gc.service
```
ymmt2005: ditto

ignitions/common/systemd/docker-prune.timer
dulltz: The intention is to stop docker-prune.service when docker-prune.timer is stopped.

Sorry, I intended to specify docker-prune.service as PartOf docker-prune.timer,
but I specified the relationship in reverse.

PR #550, opened 2019-09-18
GCP: use c2 instances instead of n1

ymmt2005:
c2 intance type performs better than n1.
It requires at least 4 local SSDs.


ymmt2005:
Pending until Google approves my C2 CPU quota increase request.

ymmt2005:
not improved.

PR #555, opened 2019-09-24
ignitions: Set HTTP(s) proxy from neco metadata

mitsutaka:
Squid service is deployed as type: LoadBalancer for using proxy if
own kube-proxy is stopped.


ymmt2005:
We discussed this and concluded this needs to be re-designed.

PR #556, opened 2019-09-25
Set proxy URL with ignitions template

mitsutaka:
Boot server wants to change `HTTP_PROXY` and `HTTPS_PROXY`.
The solution is setting proxy URL in etcd and reflecting the value with ignitions template.
After reflection it, all worker nodes need to be rebooted.

- `neco`: add `node-proxy` config for ignitions.
- ignition: Use configurable http proxy if exists.
- dctest: Refactor of using serf member status.
- squid: Change to `type: LoadBalancer`.


mitsutaka:
> What are the changes for dctest? Is it related to this issue?

This is just for refactoring code which uses serf members. I have updated the description.

PR #625, opened 2019-10-30
Add retry when initializing data for sabakan.

llamerada-jp:
Add Eventually method to retry when initializing data for sabakan.


ymmt2005:
Please retry in the logic.

PR #630, opened 2019-10-31
dctest: add admin who can login with password

dulltz:
- Replace ignition's passwd.yml if the environment is dctest
- Add `admin` user to the passwd.yml for dctest
- The password of `admin` is "admin"


dulltz:
renew branch https://github.com/cybozu-go/neco/pull/637

PR #701, opened 2019-12-17
Update components for k8s 1.16

tapih:
- Update components in `Makefile.tools`
- Enable EvenPodsSpread feature gate
    - add the option(`--feature-gates="EvenPodsSpread=true"`) to both extra_args fields of kube-apiserver and kube-scheduler in cke-template.yaml
    - https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
- Update artifacts (CKE & Coil)

### DIFF
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -53,7 +53,7 @@ containerd:
 crictl:
 	mkdir -p $(LIBEXECDIR) $(DOCDIR)/$@/
 	$(CURL) -o crictl.tar.gz https://github.com/kubernetes-sigs/cri-tools/releases/download/v$(CRITOOLS_VERSION)/crictl-v$(CRITOOLS_VERSION)-linux-amd64.tar.gz
-	tar zxf crictl.tar.gz -C $(LIBEXECDIR)
+	tar zxf crictl.tar.gz --no-same-owner -C $(LIBEXECDIR)
 	rm -f crictl.tar.gz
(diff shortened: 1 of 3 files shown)


Makefile.tools
```diff
@@ -10,24 +10,24 @@ CURL=curl -Lsf
...
+TOPOLVM_VERSION = 0.2.1
 
 ### for stern
 STERN_VERSION = 1.11.0
 
 ### for kustomize
 # Follow Argo CD installed kustomize version
 # https://github.com/cybozu/neco-containers/blob/master/argocd/Dockerfile#L32
-KUSTOMIZE_VERSION = 3.1.0
+KUSTOMIZE_VERSION = 3.2.0
```
ymmt2005: The latest version seems 3.5.3.  Why do you choose 3.2.0?
https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.5.3

etc/cke-template.yml
```diff
@@ -23,7 +23,9 @@ pod_subnet: 10.64.0.0/14
 dns_service: internet-egress/unbound
 options:
   kube-api:
-    extra_args: ["--enable-admission-plugins=PodSecurityPolicy"]
+    extra_args:
+      - --enable-admission-plugins=PodSecurityPolicy
+      - --feature-gates="EvenPodsSpread=true"
```
ymmt2005: This seems invalid as YAML is not a shell.  Probably, `"--feature-gates=EventPodsSpread=true"` is good.
(or write the list as JSON).

Makefile.tools
masa213f: This version(v3.2.0) is supported by ArgoCD.

Issue #702, opened 2019-12-18
Set up GCP instances automatically delete themselves after a set time


ysksuzuki:
## What

Currently GCP instance is removed via GAE, however the removal job is failing on some user's environments. It could lead to extra costs. 

## How

- Set a process or job that remove the instance itself with gcloud API.
- Add appropriate privilege to the instance if necessary. 

reference:
https://cloud.google.com/community/tutorials/create-a-self-deleting-virtual-machine

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #728, opened 2020-01-14
Auto shutdown GCP instance

tapih:
https://github.com/cybozu-go/neco/issues/702


gcp/compute.go
```diff
@@ -6,22 +6,39 @@ import (
...
+
+at {{ .ShutdownAt }} -f $STARTUP_PATH
+`))
+
 const (
 	retryCount   = 300
 	imageLicense = "https://www.googleapis.com/compute/v1/projects/vm-options/global/licenses/enable-vmx"
 	// MetadataKeyExtended is key for extended time in metadata
 	MetadataKeyExtended = "extended"
+	startUpScriptPath   = "/tmp/gcp-instance-startup.sh"
```
ymmt2005: Why don't you use [`ioutil.Tempfile`](https://golang.org/pkg/io/ioutil/#TempFile)?

gcp/compute.go
```diff
@@ -88,18 +105,54 @@ func (cc *ComputeClient) CreateVMXEnabledInstance(ctx context.Context) error {
...
 
+func convertLocalTimeToUTC(timezone, shutdownAt string) (string, error) {
+	loc, err := time.LoadLocation(timezone)
+	if err != nil {
+		return "", err
+	}
+	now := time.Now().In(loc)
+	localTime := fmt.Sprintf("%d-%02d-%02d "+shutdownAt+":00",
+		now.Year(), now.Month(), now.Day())
+	t, err := time.ParseInLocation(timeFormat, localTime, loc)
```
ymmt2005: Ain't it equal to the following?

```go
t, err := time.ParseInLocation("15:04", shutdownAt, loc)
```

tapih: As a result of an experiment, I found that time format also needs a year and date to recognize timezone correctly.
`time.ParseInLocation("15:04", shutdownAt, loc)` gets the output like below:

```
0000-01-01 21:00:00 +0918 LMT
```

gcp/compute.go
```diff
@@ -88,18 +104,62 @@ func (cc *ComputeClient) CreateVMXEnabledInstance(ctx context.Context) error {
...
+		ShutdownAt: shotdownAt,
+	})
+	if err != nil {
+		return err
+	}
+	tmpfile, err := ioutil.TempFile("/tmp", "gcp-start-up-script-*.sh")
+	if err != nil {
+		return err
+	}
+	log.Info("start up script for "+tmpfile.Name(), map[string]interface{}{})
```
ymmt2005: should delete the temporary file as follows:

```go
defer func() {
    tmpfile.Close()
    os.Remove(tmpfile.Name())
}()
```

gcp/compute.go
```diff
@@ -88,18 +104,62 @@ func (cc *ComputeClient) CreateVMXEnabledInstance(ctx context.Context) error {
...
+	_, err = tmpfile.Write(buf.Bytes())
+	if err != nil {
+		return err
```
ymmt2005: This leaks an open file descriptor!
Remember that open files should always be closed in a `defer` clause.

Issue #661, opened 2019-11-28
preStop hook for graceful termination


ymmt2005:
## What

For true graceful termination of Pods, containers should have preStop pod and sleep
for a while (say 2 seconds) before it starts shutting down.  Otherwise, the Pod may receive
new connections and returns service unavailable errors because it still exists in Endpoints.

https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods

Especially, Squid pod should have such a preStop hook.

## How

- Add preStop hook to wait for 5 seconds to the following pods:
  - contour
  - envoy
  - squid
  - cluster-dns
  - unbound
- cybozu-go/well: Add sleep into signal handler
- kubernetes-sigs/controller-runtime: Send PR to add sleep into signal handler

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #732, opened 2020-01-16
Sleep in preStop hook for graceful termination

tapih:
https://github.com/cybozu-go/neco/issues/661

### DIFF
--- a/etc/squid.yml
+++ b/etc/squid.yml
@@ -237,10 +237,7 @@ spec:
           lifecycle:
             preStop:
               exec:
-                command:
-                  - sh
-                  - -c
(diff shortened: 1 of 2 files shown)


etc/squid.yml
```diff
@@ -234,6 +234,13 @@ spec:
         - name: squid
           image: {{ .squid }}
           imagePullPolicy: IfNotPresent
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                  - sh
+                  - -c
+                  - "sleep 5"
```
ymmt2005: I prefer `command: ["sh", "-c", "sleep", "5"]` shortly.

PR #762, opened 2020-01-23
add setup-partition service

llamerada-jp:
Fix ignition files and a udev rule for SS nodes.
This PR gives the following change:

- Create a partition in each encrypted HDD on SS nodes.
- Create symlinks to the partitions into `/dev/crypt-part/by-path/`.
- Some refactor of CircleCI Jobs.

### DIFF
--- a/docs/ignition.md
+++ b/docs/ignition.md
@@ -179,7 +179,7 @@ There are clear and strong reasons why the process is ordered this way.
 
 * What type of devices are partitioned?
 
-    All encrypted HDDs. In addition, the created partitions have the static name (`primary`).
+    All encrypted HDDs.
 
 systemd targets and ordering dependencies
 -----------------------------------------
--- a/docs/ignition.md
+++ b/docs/ignition.md
@@ -177,9 +177,9 @@ There are clear and strong reasons why the process is ordered this way.
     - OSD should be backed by whole disk, partitions or logical volumes.
     - Making partition in dm-crypted device can bypass the above-mentioned limitation.
 
-* What devices are partitioned?
+* What type of devices are partitioned?
 
-    All encrypted HDDs. In addition, it makes persistent names for each partition.
+    All encrypted HDDs. In addition, the created partitions have the static name (`primary`).
 
 systemd targets and ordering dependencies
 -----------------------------------------
--- a/docs/ignition.md
+++ b/docs/ignition.md
@@ -43,6 +43,10 @@ The boot process runs roughly as follows:
 
 There are clear and strong reasons why the process is ordered this way.
 
+In addition to the above, the below processes run just on storage servers(`ss`):
+
+1. Prepare partitions on encrypted disks (After setup dm-crypt volumes using `sabakan-cryptsetup`)
+
 ### Configure network with DHCP
 
 * Why the network need to be configured with DHCP and later be re-configured?
@@ -161,6 +165,18 @@ There are clear and strong reasons why the process is ordered this way.
 
     Because running `serf` will tell sabakan that the machine becomes running
     and available.
+
+### Prepare partitions on encrypted disks
+
+* Why partitions are created?
+
+    Because devices on `ss` are used as storage disks for `OSD` pod of `Rook/Ceph`.
+
+* How partitions are created?
+
+    By running the  `setup-partition` as a one-shot service on `ss`.
+    Partitions are created in encrypted devices exclude BOSS or a partition existed already.
+    And aliases specific to the device are made using `udev`.
 
 systemd targets and ordering dependencies
 -----------------------------------------
--- a/docs/ignition.md
+++ b/docs/ignition.md
@@ -41,11 +41,11 @@ The boot process runs roughly as follows:
 8. Configure BIOS and BMC
 9. Run `serf` and other programs as a docker container
 
-There are clear and strong reasons why the process is ordered this way.
-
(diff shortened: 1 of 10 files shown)


ignitions/roles/ss/files/opt/sbin/setup-partition
```diff
@@ -0,0 +1,33 @@
...
+# debug
+set -x
+
+is_target_device() {
+    DEV=$(basename $1)
+    if [ $DEV = 'vda' ] || [ $DEV = 'vdb' ]; then
+        # in the case of qemu's system disks, enter here
+        echo 'false'
+    elif [ -e /sys/block/$DEV/device/model ] &&
+         grep -q DELLBOSS /sys/block/$DEV/device/model; then
```
ymmt2005: How about excluding NVMe devices?

I'd like to suggest to use a pattern match under /dev/disk/by-path of the original device
and pick up only HDDs.

masa213f: This script is placed only on SS nodes.
There is no NVMe in SS nodes. So I think, We need not check NVMe.

ymmt2005: Understood.

docs/ignition.md
```diff
@@ -162,6 +166,18 @@ There are clear and strong reasons why the process is ordered this way.
...
+* Why partitions are created?
+
+    Because devices on `ss` are used as storage disks for `OSD` pod of `Rook/Ceph`.
+
+* How partitions are created?
+
+    By running the  `setup-partition` as a one-shot service on `ss`.
+    Partitions are created in encrypted devices exclude BOSS or a partition existed already.
+    And aliases specific to the device are made using `udev`.
+
```
satoru-takeuchi: * What devices are encrypted?

This script makes a partition in all encrypted HDDs. In addition, it makes persistent names for each partitions.

docs/ignition.md
```diff
@@ -43,6 +43,10 @@ The boot process runs roughly as follows:
 
 There are clear and strong reasons why the process is ordered this way.
 
+In addition to the above, the below processes run just on storage servers(`ss`):
+
```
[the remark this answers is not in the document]
masa213f: Thanks!

docs/ignition.md
```diff
@@ -43,6 +43,10 @@ The boot process runs roughly as follows:
...
+1. Prepare partitions on encrypted disks (After setup dm-crypt volumes using `sabakan-cryptsetup`)
+
```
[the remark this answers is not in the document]
masa213f: "setup dm-crypt volumes (sabakan-cryptsetup)" and "prepare partition" are other steps.
I think the current explanation is more accurate.

docs/ignition.md
```diff
@@ -162,6 +166,18 @@ There are clear and strong reasons why the process is ordered this way.
     Because running `serf` will tell sabakan that the machine becomes running
     and available.
 
+### Prepare partitions on encrypted disks
+
+* Why partitions are created?
```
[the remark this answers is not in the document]
masa213f: Thanks!

docs/ignition.md
```diff
@@ -162,6 +166,18 @@ There are clear and strong reasons why the process is ordered this way.
...
+
+    Because devices on `ss` are used as storage disks for `OSD` pod of `Rook/Ceph`.
```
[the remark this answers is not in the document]
masa213f: Thanks!

docs/ignition.md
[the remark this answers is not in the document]
masa213f: > Why we need to do it is explained in the next section.

I skipped it. Because there was already exist navigation to the next section.

docs/ignition.md
masa213f: Thanks! But I think we need not describe `script` here.
Because the `script` is an internal implementation.

docs/ignition.md
```diff
@@ -162,6 +166,21 @@ There are clear and strong reasons why the process is ordered this way.
...
+* Why partitions should be created?
+
+    Because of the following reasons.
+
+    - The devices on `ss` are used as OSD of Rook/Ceph
+    - These devices are encrypted with dm-crypt.
+    - OSD should be backed by whole disk, partitions or logical volumes.
+    - Making partition in dm-crypted device can bypass the above-mentioned limitation.
+
+* What devices are partitioned?
```
ymmt2005: What type of devices

docs/ignition.md
```diff
@@ -162,6 +166,21 @@ There are clear and strong reasons why the process is ordered this way.
...
+
+    All encrypted HDDs. In addition, it makes persistent names for each partition.
```
ymmt2005: > In addition, it makes persistent names for each partition.

adding to _what_??  I cannot understand this sentence.

docs/ignition.md
masa213f: Thanks!

docs/ignition.md
masa213f: Thanks, it was strange. I updated as follows. How is this? 

> In addition, the created partitions have the static name (`primary`).

ymmt2005: I think "In addition," does not make sense.

Just remove the second sentence.  If you want to address "persistent names", say it in another section.

masa213f: Thanks. I deleted the sentence.

Issue #704, opened 2019-12-19
Kill old login sessions on the boot servers


masa213f:
## What

The old login session may remain on the boot server in stage and prod.

```
cybozu@stage0-boot-0:~$ w
 02:10:14 up 153 days, 18:50,  9 users,  load average: 0.08, 0.52, 0.56
USER     TTY      FROM             LOGIN@   IDLE   JCPU   PCPU WHAT
cybozu   pts/12   tmux(11750).%0   28Aug19 111days  0.20s  0.20s -bash
cybozu   pts/13   tmux(11750).%1   28Aug19 111days  2:27m  2:27m watch kubectl get certificates,dnsendpoint,ingressroute -n monitoring
```

## How

- Create a script to kill old login sessions
- Run it regularly by crontab.
- [`loginctl`](http://manpages.ubuntu.com/manpages/bionic/man1/loginctl.1.html) can be used to identify and terminate old user sessions.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #767, opened 2020-01-29
Kill old login sessions on the boot servers

tapih:
https://github.com/cybozu-go/neco/issues/704


debian/usr/sbin/kill-old-login-sessions
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh -e
+
+limit=$(echo "3600 * 24 * 1" | bc) # 1 day
```
ymmt2005: I think `limit=86400` is just enough, ain't it?

PR #775, opened 2020-02-04
Fix setup-partition

masa213f:
Create partitions using `gdisk` instead of `parted`.
The partitions created by parted isn't recognized properly. It would come from a bug in parted
that can't handle 4KB sector properly. Let's use gdisk and kpartx instead.

### DIFF
--- a/ignitions/roles/ss/files/opt/sbin/setup-partition
+++ b/ignitions/roles/ss/files/opt/sbin/setup-partition
@@ -1,6 +1,27 @@
 #!/bin/sh -u
 
-PART_NAME=primary
+PART_TYPE=f800
+PART_NAME=ceph-osd
+
+# Execute the following subcommands.
+#  1. o : create a new empty GUID partition table (GPT)
+#  2. n : add a new partition
+#  3. c : change a partition's name
+#  4. w : write table to disk and exit
+GDISK_CMD=$(cat << EOS
+o
+y
+n
+
+
+
+$PART_TYPE
+c
+$PART_NAME
+w
+y
+EOS
+)
 
 is_target_device() {
     DEV=$(basename $1)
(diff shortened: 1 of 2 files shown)


satoru-takeuchi:
> Create partitions using gdisk instead of parted.
> Because, on the actual SS nodes, the partitions created by parted ware not recognized by DeviceMapper properly.

To make the problem clearer, how about the following commit description?

```
Create partitions ....
The partitions created by parted isn't recognized properly. It would come from a bug in parted
that can't handle 4KB sector properly. Let's use gdisk and kpartx instead.
```

ignitions/roles/ss/files/opt/sbin/setup-partition
```diff
@@ -1,6 +1,27 @@
 #!/bin/sh -u
 
-PART_NAME=primary
+PART_TYPE=f800
+PART_NAME=ceph-osd
+
+# Execute the following subcommands.
```
satoru-takeuchi: The following description seems to be the command reference. We can't understand what GDISK_CMD actually does from this. How about what you want to do here instead? Like "The following command creates a partition. Its type is $PART_TYPE and its name is $PART_NAME".

ignitions/roles/ss/files/opt/sbin/setup-partition
```diff
@@ -20,13 +41,13 @@ for DM in $(ls /dev/dm-*); do
     DEV_PATH=$(cryptsetup status $DM | grep device | awk '{print $2}')
     if [ -n "$DEV_PATH" ]; then
         if [ $(is_target_device $DEV_PATH) = 'true' ]; then
-            if [ -n "$(parted -s $DM print 2> /dev/null | grep $PART_NAME)" ]; then
-                echo "partition is already exists on $DM (backed by $DEV_PATH)"
-                partprobe $DM
-            else
+            tmp=$(echo i | gdisk $DM 2> /dev/null | grep "Partition name:" | grep "'$PART_NAME'")
+            if [ -z "$tmp" ]; then
```
satoru-takeuchi: FYI, although you can't change your code here. `grep -q` is useful in this kind of situation. It returns 0 if match and returns 1 if not match, without any output.

ignitions/roles/ss/files/opt/sbin/setup-partition
```diff
@@ -20,13 +41,13 @@ for DM in $(ls /dev/dm-*); do
     DEV_PATH=$(cryptsetup status $DM | grep device | awk '{print $2}')
     if [ -n "$DEV_PATH" ]; then
         if [ $(is_target_device $DEV_PATH) = 'true' ]; then
```
ymmt2005: The following can make the code more readable:

```shell
for DM in $(ls /dev/dm-*); do
    DEV_PATH=$(cryptsetup status $DM | grep device | awk '{print $2}')
    if [ -n "$DEV_PATH" ]; then
        if [ $(is_target_device $DEV_PATH) != true ]; then continue; fi
        tmp=$(echo i | gdisk $DM 2> /dev/null | grep "Partition name:" | grep "'$PART_NAME'")
        # snip
    fi
done    
```

satoru-takeuchi:
LGTM

PR #872, opened 2020-03-23
Modify link target from part to crypt-disk.

kfyharukz:


kfyharukz:
Close because of rename branch.

PR #873, opened 2020-03-24
Modify link target from part to crypt-disk.

kfyharukz:
This PR changes scripts for creating partitions and symbolic links on storage servers.
The targets of symbolic links are changed from partitions to crypt-disks.
These links are pointed from PersistentVolumes, which are finally used by Rook/Ceph.

Partitions are not created any more.


ignitions/roles/ss/site.yml
```diff
@@ -2,7 +2,3 @@ include: ../../common/common.yml
 files:
   - /etc/udev/crypt-base-path
   - /etc/udev/rules.d/99-neco.rules
-  - /opt/sbin/setup-partition
-systemd:
-  - name: setup-partition.service
```
masa213f: Please delete the following file.
https://github.com/cybozu-go/neco/blob/2def6024c2f181e695beebff1ddfda2e395d8ffb/ignitions/roles/ss/systemd/setup-partition.service

ignitions/roles/ss/files/etc/udev/rules.d/99-neco.rules
```diff
@@ -1,6 +1,7 @@
 KERNEL!="dm-*", GOTO="neco_end"
-ENV{ID_PART_ENTRY_NAME}!="ceph-osd", GOTO="neco_end"
+ENV{DM_NAME}=="crypt-vda|crypt-vdb", GOTO="neco_end"
+ENV{DM_NAME}!="crypt-*", GOTO="neco_end"
```
masa213f: Personally, this order is easier to understand. (overall condition -> exception)
This is a personal feeling. So you don't have to change it if you don't mind.

```suggestion
ENV{DM_NAME}!="crypt-*", GOTO="neco_end"
# Skip the qemu's system disks
ENV{DM_NAME}=="crypt-vda|crypt-vdb", GOTO="neco_end"
```

PR #906, opened 2020-04-09
add deb package build for operator tools

Hsn723:
This adds the creation of a `neco-operation-cli` package containing a subset of what's bundled in the neco deb, namely tools oriented towards operator-use and their documentation/license files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,7 @@ $(OP_DEB):
 		cp $(BINDIR)/$$BINNAME $(OPBINDIR) ; \
 		cp -r $(DOCDIR)/$$BINNAME $(OPDOCDIR) ; \
 	done
+	$(MAKE) -f Makefile.tools SUDO=$(SUDO) BINDIR=$(OPBINDIR) DOCDIR=$(OPDOCDIR) teleport
 	$(FAKEROOT) dpkg-deb --build $(DEBBUILD_FLAGS) $(OPWORKDIR) $(DEST)
 
 gcp-deb: setup-files-for-deb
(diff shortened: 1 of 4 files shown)


tapih:
I was also working on [the task](https://github.com/cybozu-go/neco/pull/909) to compress the linux binaries into a zip file. However, It's better to install them with a deb package because it's slightly easier to install the binaries.

Just to confirm, do you assume that the deb package is uploaded to artifactory to let you install the binaries in the production environment?

Makefile
```diff
@@ -78,6 +79,15 @@ setup-files-for-deb:
 $(DEB): setup-files-for-deb
 	$(FAKEROOT) dpkg-deb --build $(DEBBUILD_FLAGS) $(WORKDIR) $(DEST)
 
+$(OP_DEB):
+	mkdir -p $(OPBINDIR) $(OPDOCDIR) $(OPWORKDIR)/DEBIAN
+	sed 's/@VERSION@/$(patsubst v%,%,$(VERSION))/; /Package: neco/s/$$/-operation-cli/; s/Continuous delivery tool/Operation tools/' debian/DEBIAN/control > $(OPCONTROL)
+	for BINNAME in argocd kubectl kustomize stern; do \
```
tapih, line 85: Please add teleport., which is downloaded [here](https://get.gravitational.com/teleport-v4.2.8-linux-amd64-bin.tar.gz).

Hsn723:
@tapih 
I've added teleport. Based on the Windows binaries, I'm assuming we only need `tsh`.

> Just to confirm, do you assume that the deb package is uploaded to artifactory to let you install the binaries in the production environment?

Yes, I'm assuming that the generated deb is going to be (manually?) uploaded to artifactory similarly to the `operation-cli-windows` zip.

llamerada-jp:
merged by following PR.
https://github.com/cybozu-go/neco/pull/914

Issue #903, opened 2020-04-08
Release ArgoCD windows binary and some ubuntu binaries


masa213f:
## What

Release the following commands in the `cybozu-go/neco`.

- argocd windows binary. 
  - Release it in the existing zip file (`operation-cli-windows-amd64.zip`).
- linux(ubuntu) binaries of `argocd`, `kubectl` and `tsh`
  - Release in the new zip file (e.g. `operation-cli-linux-amd64.zip`).

## How

Update the following program.
https://github.com/cybozu-go/neco/blob/master/pkg/download-binaries/main.go

Also, update [the operation template](https://github.com/cybozu-private/neco-ops/blob/master/.github/ISSUE_TEMPLATE/prod-neco.md) to upload released ZIP archives to artifactory.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #909, opened 2020-04-10
Download windows binaries in Makefile

tapih:
Implements https://github.com/cybozu-go/neco/issues/903.

Windows binaries are now downloaded in ./pkg/download_binaries.go.
This process is called only in `test` and `build_release` workflows, and they do not need to download binaries quickly.
And now the versions of the binaries are fetched over the Internet, but this should be changed to avoid downloading unexpected versions.

So, in this PR, move downloading logic from go code to Makefile and use the same binary versions with the ones of linux.


Makefile.tools
```diff
@@ -20,7 +20,6 @@ K8S_VERSION = 1.17.4
 
 ### for lvmd
 TOPOLVM_VERSION = 0.4.3
-
```
masa213f, line 23: Don't delete this blank line.

Makefile.tools
```diff
@@ -62,18 +61,24 @@ crictl:
 	$(CURL) -o $(DOCDIR)/$@/README.md https://raw.githubusercontent.com/kubernetes-sigs/cri-tools/v$(CRITOOLS_VERSION)/README.md
 
 argocd:
-	mkdir -p $(BINDIR) $(DOCDIR)/$@/
+	mkdir -p $(BINDIR) $(DOCDIR)/$@/ $(WINDOWS_BINDIR)/$@/ $(WINDOWS_DOCDIR)/$@/
 	$(CURL) -o $(BINDIR)/argocd https://github.com/argoproj/argo-cd/releases/download/v$(ARGOCD_VERSION)/argocd-linux-amd64
 	chmod +x $(BINDIR)/argocd
 	$(CURL) -o $(DOCDIR)/$@/LICENSE https://raw.githubusercontent.com/argoproj/argo-cd/v$(ARGOCD_VERSION)/LICENSE
 	$(CURL) -o $(DOCDIR)/$@/README.md https://raw.githubusercontent.com/argoproj/argo-cd/v$(ARGOCD_VERSION)/README.md
+	$(CURL) -o $(WINDOWS_BINDIR)/$@/argocd https://github.com/argoproj/argo-cd/releases/download/v$(ARGOCD_VERSION)/argocd-windows-amd64.exe
```
masa213f, line 69: The extension(`.exe`) is required.

```suggestion
	$(CURL) -o $(WINDOWS_BINDIR)/$@/argocd.exe https://github.com/argoproj/argo-cd/releases/download/v$(ARGOCD_VERSION)/argocd-windows-amd64.exe
```

Makefile.tools
```diff
@@ -94,11 +99,16 @@ kustomize:
 	$(CURL) -o $(DOCDIR)/$@/README.md https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v${KUSTOMIZE_VERSION}/README.md
 
 teleport:
-	mkdir -p $(BINDIR) $(DOCDIR)/$@/
+	$(SUDO) apt-get update
+	$(SUDO) apt-get -y install --no-install-recommends unzip
+	mkdir -p $(BINDIR) $(DOCDIR)/$@/ $(OPDOCDIR)/$@/ $(WINDOWS_BINDIR)/$@/ $(WINDOWS_DOCDIR)/$@/
```
masa213f, line 104: `$(OPDOCDIR)` should not be used here.
Because `$(OPDOCDIR)/teleport` will be overwritten at the following target.
https://github.com/cybozu-go/neco/blob/2648dd2adc83991a401d119d1d33c5a26143d080/Makefile#L87

Makefile.tools
```diff
@@ -94,11 +99,16 @@ kustomize:
 	$(CURL) -o $(DOCDIR)/$@/README.md https://raw.githubusercontent.com/kubernetes-sigs/kustomize/kustomize/v${KUSTOMIZE_VERSION}/README.md
 
 teleport:
-	mkdir -p $(BINDIR) $(DOCDIR)/$@/
+	$(SUDO) apt-get update
+	$(SUDO) apt-get -y install --no-install-recommends unzip
```
masa213f, line 103: Don't install an extra package here.
The extra package should be added to `PACKAGES`.
https://github.com/cybozu-go/neco/blob/2648dd2adc83991a401d119d1d33c5a26143d080/Makefile#L15

.circleci/config.yml
```diff
@@ -120,9 +121,7 @@ jobs:
           command: |
             VERSION=$(echo $CIRCLE_TAG | sed 's/^release-//')
             time make deb SUDO="" DEBBUILD_FLAGS="" VERSION="$VERSION" TAGS=release FAKEROOT=
-      - run:
-          name: download binaries to workspace
-          command: go run ./pkg/download-binaries/main.go
+            time make tools SUDO="" DEBBUILD_FLAGS="" VERSION="$VERSION" TAGS=release FAKEROOT=
```
masa213f, line 124: `make tools` also create a zip file.
There is a contradiction between the name("Build deb package") and the logic.

So it is better to split the run section.

PR #996, opened 2020-06-03
update cke to v1.17.8

masa213f:


masa213f:
CKE 1.17.8 is cancelled due to the following bug.
https://github.com/cybozu-go/cke/pull/348

PR #1004, opened 2020-06-08
Periodical update of components.

yokaze:
This PR is for the periodical update of components.
ref: https://github.com/cybozu-private/neco-ops/issues/405


ymmt2005:
@zoetrope 
IIRC, system containers are `rkt` images that are pre-installed on boot servers.
They cannot be updated.

kfyharukz:
@morimoto-cybozu I think those components are highly related with Kubernetes, so they should be updated in a Kubernetes updating task, shouldn't it?

PR #1013, opened 2020-06-15
Update lvmd for TopoLVM v0.5.0

zoetrope:


ignitions/roles/cs/site.yml
```diff
@@ -1,6 +1,7 @@
 include: ../../common/common.yml
 files:
   - /opt/sbin/setup-lvmd
+  - /etc/lvmd/lvmd.yaml
```
ymmt2005, line 4:
```suggestion
  - /etc/topolvm/lvmd.yaml
```

ignitions/roles/cs/files/etc/topolvm/lvmd.yaml
```diff
@@ -0,0 +1,5 @@
+device-classes:
+  - name: "ssd"
+    volume-group: "vg1"
+    default: true
+    spare-gb: 1
```
ymmt2005, line 5: We didn't specify this value, so let's omit this.
```suggestion
```

PR #1024, opened 2020-06-23
ignition: exec udevadm trigger periodically

masa213f:

### DIFF
--- a/ignitions/common/systemd/udev-trigger.timer
+++ b/ignitions/common/systemd/udev-trigger.timer
@@ -1,11 +1,13 @@
 [Unit]
 Description=Request udev device events periodically
+DefaultDependencies=no
 Wants=systemd-udevd.service
 After=systemd-udevd-kernel.socket systemd-udevd-control.socket
 Before=setup-var.service
 
 [Timer]
-OnCalendar=*-*-* *:*:00
+OnActiveSec=60
+OnUnitActiveSec=60
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


ignitions/common/systemd/udev-trigger.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Request udev device events periodically
+Wants=systemd-udevd.service
+After=systemd-udevd-kernel.socket systemd-udevd-control.socket
+Before=setup-var.service
+
+[Timer]
+OnCalendar=*-*-* *:*:00
```
ymmt2005, line 8: I think it is better to use monolithic locks.
```suggestion
OnActiveSec=60
```

ignitions/common/systemd/udev-trigger.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Request udev device events periodically
```
ymmt2005, line 2: Add `DefaultDependencies=no`.  This is necessary to run this unit before `setup-var.service`.

ignitions/common/systemd/udev-trigger.timer
```diff
@@ -0,0 +1,12 @@
...
+DefaultDependencies=no
+Wants=systemd-udevd.service
+After=systemd-udevd-kernel.socket systemd-udevd-control.socket
+Before=setup-var.service
+
+[Timer]
+OnActiveSec=60
```
ymmt2005, line 9: To reset the timer after each service invocation, we should add `OnUnitActiveSec=` instead.
```suggestion
OnActiveSec=60
OnUnitActiveSec=60
```

PR #1053, opened 2020-07-09
Make GAE app delete test instances both from asia-northeast1-c and asia-northeast2-c

yokaze:

### DIFF
--- a/gcp/app/server.go
+++ b/gcp/app/server.go
@@ -52,7 +52,8 @@ func (s Server) HandleShutdown(w http.ResponseWriter, r *http.Request) {
 
 func (s Server) shutdown(w http.ResponseWriter, r *http.Request) {
 	project := s.cfg.Common.Project
-	zone := s.cfg.Common.Zone
+	commonZone := s.cfg.Common.Zone
+	addZones := s.cfg.App.Shutdown.AdditionalZones
 	exclude := s.cfg.App.Shutdown.Exclude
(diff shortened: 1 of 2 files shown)


gcp/app/server.go
```diff
@@ -64,60 +65,64 @@ func (s Server) shutdown(w http.ResponseWriter, r *http.Request) {
...
 
-	for _, instance := range instanceList.Items {
-		if contain(instance.Name, exclude) {
-			continue
+	targetZones := append([]string{commonZone}, addZones...)
+	for _, zone := range targetZones {
+		instanceList, err := service.Instances.List(project, zone).Do()
+		if err != nil {
+			RenderError(r.Context(), w, InternalServerError(err))
+			return
```
zoetrope, line 74: It might be better to change from `return` to `continue`.

gcp/app/server.go
```diff
@@ -64,60 +65,64 @@ func (s Server) shutdown(w http.ResponseWriter, r *http.Request) {
...
+		for _, instance := range instanceList.Items {
+			if contain(instance.Name, exclude) {
+				continue
+			}
+
+			shutdownAt, err := getShutdownAt(instance)
+			if err != nil {
+				if err != errShutdownMetadataNotFound {
+					RenderError(r.Context(), w, InternalServerError(err))
+					return
```
zoetrope, line 86: ditto

PR #1085, opened 2020-07-29
Use CircleCI cache for downloaded artifacts

yokaze:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,23 @@ commands:
             if [ -f .skip ]; then exit 0; fi
             ./bin/run-dctest.sh bootstrap "<<parameters.tag>>" "<<parameters.datacenter>>"
           no_output_timeout: 20m
+  setup-tools:
+    description: "setup tools"
+    steps:
+      # CircleCI has not yet supported to use environment variables in cache keys, so we need to use a workaround.
+      # https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache
+      - run: echo $CACHE_VERSION > CACHE_VERSION
+      - restore_cache:
+          key: setup-tools-{{ checksum "CACHE_VERSION" }}-{{ checksum "Makefile" }}-{{ checksum "Makefile.tools" }}
+      - run:
+          name: Download tools if needed
+          command:
+            if [ ! -d work ]; then time make setup-tools; fi
+      - save_cache:
+          key: setup-tools-{{ checksum "CACHE_VERSION" }}-{{ checksum "Makefile" }}-{{ checksum "Makefile.tools" }}
+          paths:
+            - build
+            - work
   notify-slack-to-extend:
     description: "notify slack to extend the lifetime of the GCP instance"
     steps:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -56,6 +56,23 @@ commands:
             if [ -f .skip ]; then exit 0; fi
             ./bin/run-dctest.sh bootstrap "<<parameters.tag>>" "<<parameters.datacenter>>"
           no_output_timeout: 20m
+  setup-tools:
+    description: "setup tools"
+    steps:
+      # CircleCI has not yet supported to use environment variables in cache keys, so we need to use a workaround.
+      # https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache
+      - run: echo $CACHE_VERSION > CACHE_VERSION
```
ymmt2005, line 64: What is `CACHE_VERSION` and why do you use this as a checksum?

Makefile
```diff
@@ -69,9 +69,12 @@ deb: $(DEB)
...
 
 setup-tools:
-	$(MAKE) -f Makefile.tools SUDO=$(SUDO)
+	$(MAKE) -f Makefile.tools
 
-setup-files-for-deb: setup-tools
+check-tools:
+	$(MAKE) -f Makefile.tools check-all
+
+setup-files-for-deb: check-tools
```
ymmt2005, line 77: Why don't you just depend on `setup-tools`?

Makefile
```diff
@@ -93,11 +96,10 @@ $(OP_DEB): setup-files-for-deb
 		cp $(BINDIR)/$$BINNAME $(OPBINDIR) ; \
 		cp -r $(DOCDIR)/$$BINNAME $(OPDOCDIR) ; \
 	done
-	$(MAKE) -f Makefile.tools SUDO=$(SUDO) BINDIR=$(OPBINDIR) DOCDIR=$(OPDOCDIR) teleport
 	$(FAKEROOT) dpkg-deb --build $(DEBBUILD_FLAGS) $(OPWORKDIR) $(DEST)
 
-$(OP_ZIP): setup-tools
-	cd $(WORKDIR)/windows/ && zip -r $(abspath .)/$@ *
+$(OP_ZIP): check-tools
```
ymmt2005, line 101: ditto

.circleci/config.yml
yokaze: It is discussed [here](https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache).

- `CACHE_VERSION` is a recommended variable from CircleCI that allows us to clear corrupted caches without committing anything to GitHub. It is set in CircleCI project settings.
The value is first set as `v1` and is incremented as `v2`, `v3` if needed.

- `{{ .Environment.CACHE_VERSION }}` does not work properly now, so we need a workaround.

- As described [here](https://circleci.com/docs/2.0/caching/#using-keys-and-templates), only checksum of the file content is available in cache key.

ymmt2005: I see.

Makefile
yokaze: This is because it is difficult for the targets in [`Makefile.tools`](https://github.com/cybozu-go/neco/blob/use-circleci-cache/Makefile.tools) to satisfy the followings simultaneously.
- Ensure preparing the specified component version regardless of cache existence 
- Skip download if component version is unchanged

Though it can be done, it makes `Makefile` a bit complex.

I'll be welcome to change implementation if it is desirable.

ymmt2005: So, what will occur when `check-tools` fails?

PR #1090, opened 2020-08-04
Migrate gcp related code to github.com/cybozu-go/neco-gcp

tapih:
Signed-off-by: Hiroshi Muraoka <h.muraoka714@gmail.com>

PR #1093, opened 2020-08-10
necogcp: add snapshot commands

binoue:
`necogcp` is currently using the `home` device as a persistent volume.
However, `necogcp` does not have a function to create/restore the volume from snapshot and this cause problems when the volume is corrupt.

Therefore, this PR adds `create-snapshot` and `restore-snapshot` subcommands for handling the problem above.
The new subcommands are manually tested. See [this](https://bozuman.cybozu.com/k/#/space/3801/thread/44932/2467026/3711787)

Signed-off-by: binoue <banji-inoue@cybozu.co.jp>

### DIFF
--- a/docs/gcp/necogcp.md
+++ b/docs/gcp/necogcp.md
@@ -30,6 +30,14 @@ Synopsis
 
     Setup `host-vm` or `vmx-enabled` instance. It can run on only them.
 
+* `necogcp create-snapshott`
+
+    Create `home` volume snapshot.
+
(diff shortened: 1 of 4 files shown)


ymmt2005:
`necogcp` is going to be deleted in #1090 .
Please postpone this after it is migrated to github.com/cybozu-go/neco-gcp

ymmt2005:
@tapih Please consider picking this in neco-gcp repo.

PR #1100, opened 2020-08-31
send SIGKILL on loginctl kill-session

umezawatakeshi:
- send SIGKILL on `loginctl kill-session`
- ~kill orphaned bash processes (whose PPID is 1)~

### DIFF
--- a/debian/usr/sbin/kill-old-login-sessions
+++ b/debian/usr/sbin/kill-old-login-sessions
@@ -1,18 +1,30 @@
 #!/bin/sh -e
 
 limit=86400 # 1 day
-sessions=$(loginctl list-sessions --no-legend | sed -e 's/^ *//g' | cut -f 1 -d ' ')
+now=$(date +%s)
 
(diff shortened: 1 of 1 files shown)


debian/usr/sbin/kill-old-login-sessions
```diff
@@ -1,18 +1,30 @@
...
   if [ $diff -gt $limit ]; then
     echo "kill session:"
     loginctl show-session $session
     echo
-    loginctl kill-session $session
+    loginctl kill-session --signal 9 $session
+  fi
+done
+
+# kill orphaned bash
```
ymmt2005, line 20: @umezawatakeshi 
Have you seen orphaned bash processes after `loginctl kill-session --signal 9`?

umezawatakeshi: As I test now, orphaned bash processes are not seen. So, this block is not necessary.

Some existing orphaned bash processes on some boot servers should be deleted manually when we enable pam_teleport.

ymmt2005: I see.

ymmt2005 (approved):
LGTM

PR #1134, opened 2020-09-30
Return to asia-northeast1-c with C2 instance & NVMe local SSD

ymmt2005:


toshipp (commented):
Looks good.
BTW, I don't understand why multiple local-ssd is needed.

I think commit comment "hoge" should be changed...

ymmt2005:
@toshipp 
>  why multiple local-ssd is needed.

C2 instances allows 0, 4, or 8 local SSDs only.

toshipp:
I got it.

Would you revise commit message if other reviewers have no objection?

ymmt2005:
sure.

ysksuzuki (approved):
LGTM

PR #1192, opened 2020-11-06
add labels to VM instance

masa213f:


bin/env
```diff
@@ -6,3 +6,6 @@ MACHINE_TYPE=c2-standard-30
 DISK_TYPE=pd-ssd
 BOOT_DISK_SIZE=20GB
 GCLOUD="gcloud --quiet --account ${SERVICE_ACCOUNT} --project ${PROJECT}"
+LABEL_REPO=repo=${CIRCLE_PROJECT_REPONAME}
+LABEL_REF=ref=${CIRCLE_BRANCH}${CIRCLE_TAG}
```
zoetrope, line 10: Do you not need separator?
It would be like `masterv1.2.3`.

masa213f: I don't need a separator. The value is set only on one side of either CIRCLE_BRANCH or CIRCLE_TAG.

zoetrope: 😄

zoetrope (approved):

PR #1207, opened 2020-11-12
tuning memory size to avoid the error when CI

llamerada-jp:


toshipp:
branch renamed.
new PR is https://github.com/cybozu-go/neco/pull/1213

PR #1213, opened 2020-11-16
tuning memory size to avoid the error when CI

toshipp:
Sometimes, the CI test failed due to a kernel soft lock up crash.
It seems the linux kernel could not allocate enough memory in the watchdog timer interval.
This decreases VM memory size to reduce the possibility of such situation.


toshipp:
@tapih I wrote a rationale for the PR. Please review it.

tapih (approved):
LGTM

PR #1214, opened 2020-11-17
Ignore gcloud error when adding canceled labels

masa213f:
If an instance is deleted after listing, this gcloud fails. Prevents CI Error from this error.


tapih (approved):
LGTM

PR #1216, opened 2020-11-18
Upgrade CKE to v1.18.3

yokaze:


ymmt2005:
CKE was updated in #1217

PR #1217, opened 2020-11-19
Update coil to 2.0.3 and remove "tuning" CNI plugin

ymmt2005:
Now that Coil detects the optimal MTU value automatically,
tuning plugin is no longer necessary on GCP environment.


ysksuzuki (approved):
LGTM

PR #1224, opened 2020-11-25
Run CKE as a normal process

ymmt2005:
This PR makes CKE run without a container and removes dependency on CKE from `go.mod`.
This change is necessary to enable CKE's graceful node reboot feature.


worker/cke.go
```diff
@@ -26,14 +26,6 @@ func (o *operator) UpdateCKE(ctx context.Context, req *neco.UpdateRequest) error
...
 	if need {
-		err = cke.InstallToolsCKE(ctx)
-		if err != nil {
-			return err
-		}
-		err = cke.InstallBashCompletion(ctx)
-		if err != nil {
-			return err
-		}
 		err = o.storage.RecordContainerTag(ctx, o.mylrn, "cke")
```
ymmt2005, line 29: CKE is no longer installed as a container.

debian/DEBIAN/postinst
```diff
@@ -1,9 +1,12 @@
 #!/bin/sh -e
 
-SERVICES="node-exporter neco-updater neco-worker sabakan-state-setter ingress-watcher"
+SERVICES="node-exporter neco-updater neco-worker sabakan-state-setter ingress-watcher cke"
 TIMERS="rkt-gc kill-old-login-sessions backup-cke-etcd"
 
 configure() {
+    # to upgrade from container version of CKE
+    rm -f /etc/systemd/system/cke.service
```
ymmt2005, line 8: @yokaze 
We need to stop containerized CKE in the previous installation, don't we?
```
systemctl stop cke.service || true
```

dctest/upgrade.go
```diff
@@ -407,6 +413,27 @@ func checkVersionByDocker(address, name, image string) error {
 	return nil
 }
 
+func checkVersionOfCKE(host, image string) error {
```
ymmt2005, line 416: I don't think this is necessary because now CKE is installed as a part of the Debian package.

yokaze (approved):
Approved by @ymmt2005

PR #1230, opened 2020-11-26
Resize ss node memory size

binoue:
Currently, the ss node has only 4G memory, which is not enough to have two OSD pods. 
So, the ss node's memory is returned to 10G.


llamerada-jp (dismissed):
mobbed

llamerada-jp (dismissed):
mobbed

llamerada-jp (approved):
mobbed

PR #1233, opened 2020-11-27
Fix creation logic of deb package to include tsh binary

tapih:
The `tsh` binary for linux is not contained in the CLI deb package.
This PR fixes the code to add the binary to the package.

Signed-off-by: H.Muraoka <hiroshi-muraoka@cybozu.co.jp>


ymmt2005 (approved):
LGTM

PR #1237, opened 2020-11-30
use shell script loop instead of Acquire::Retries for apt to add interval

llamerada-jp:
Solving the problem:
- Failing apt-get command in auto-config if the connection to proxy is not enabled before exec the command.


ymmt2005:
@llamerada-jp 
What problem does this PR try to resolve?
Please elaborate on the body of PR.

llamerada-jp:
canceled

PR #1238, opened 2020-12-01
ingress-watcher: set timeout for pushing metrics

zoetrope:
`ingress-watcher` get stuck when the connection to push-gateway is refused.
This is because the default timeout setting is 0, which means no timeout on the client side.
This PR fixes this behavior. 


tapih (dismissed):
mobbed

kfyharukz (approved):
LGTM

PR #1264, opened 2020-12-10
extend /var/lib/k8s-containerd to avoid disk pressure when testing rook/ceph

llamerada-jp:


morimoto-cybozu (approved):
paired

PR #1268, opened 2020-12-11
Get rid of quay.io/cybozu/teleport

ymmt2005:
Bundle `teleport` and `tctl` binaries in the debian package and
get rid of the use of the teleport container image.


kfyharukz (approved):
LGTM

PR #1278, opened 2020-12-16
add document for MENU_ARG option

toshipp:
Signed-off-by: Toshikuni Fukaya <toshikuni-fukaya@cybozu.co.jp>

### DIFF
--- a/docs/dctest.md
+++ b/docs/dctest.md
@@ -124,13 +124,20 @@ Specify `TAGS=release` in the release branch to use `artifacts_release.go`.
 
 `make test` accepts this variable.
 
+### `MENU_ARG`
+
+Virtual data center is configured by the specified `MENU_ARG` file.
+If you change the parameter from previously used one, you should run `make clean` to apply the configuration properly.
+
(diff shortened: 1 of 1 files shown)


docs/dctest.md
```diff
@@ -124,13 +124,20 @@ Specify `TAGS=release` in the release branch to use `artifacts_release.go`.
 
 `make test` accepts this variable.
 
+### `MENU_ARG`
+
+Virtual data center is configured by the specified `MENU_ARG` file.
```
zoetrope, line 129: I don't think this explanation will help me understand what `MENU_ARG` is and why to change it.
How about introducing the following link and the purpose of changing the server configuration in the data center?
https://github.com/cybozu-go/neco/blob/master/docs/placemat-menu.md

toshipp: Ok, I will describe in more detail and add a link.

zoetrope (dismissed):

ymmt2005 (approved):
LGTM

Issue #988, opened 2020-05-28
CKE/neco: Enable Scheduling Profiles for Kubernetes v1.18 and later


masa213f:
## What

For Kubernetes v1.18 and later, Enable Scheduling Profiles to properly distribute the Pod.

https://kubernetes.io/docs/reference/scheduling/profiles/

## How

- Modify CKE to pass configuration for Scheduling Profiles.
   - It's should support [multiple profiles](https://kubernetes.io/docs/reference/scheduling/profiles/#multiple-profiles), so we need to use `v1alpha2` API.
- Set the weight as follows in neco, so that `PodTopologySpread` is reflected properly.

```
apiVersion: kubescheduler.config.k8s.io/v1alpha2
kind: KubeSchedulerConfiguration
leaderElection:
  leaderElect: true
clientConnection:
  kubeconfig: /etc/kubernetes/scheduler.conf
profiles:
  - schedulerName: default-scheduler
    plugins:
      score:
        disabled:
        - name: PodTopologySpread
        enabled:
        - name: PodTopologySpread
          weight: 500
```

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1284, opened 2020-12-18
Set PodTopologySpread weight

tapih:
Configure https://github.com/cybozu-go/neco/issues/988 .
Signed-off-by: H.Muraoka <hiroshi-muraoka@cybozu.co.jp>


binoue (approved):
LGTM

PR #1292, opened 2020-12-23
Introduce neco.ContainerRuntime interface

ymmt2005:
This PR adds `neco.ContainerRuntime`, an interface to manage containers on boot servers.
All rkt functions are migrated to this new interface.


ysksuzuki (approved):
LGTM

PR #1297, opened 2020-12-24
Refactor and cleanup

ymmt2005:
`bird` and `chrony` are managed by `artifacts.go` just the same as others.

`dctest/secrets` was a quay.io token but is no longer necessary.


ysksuzuki (approved):
LGTM

PR #1307, opened 2021-01-06
installer: purge fwupd

ymmt2005:
fwupd tries to download firmware over the Internet.
This is not possible on boot servers, so we purge it.


tapih (approved):
LGTM

PR #1352, opened 2021-01-21
Enable worker nodes to use domain names without local suffix.

morimoto-cybozu:


zoetrope (approved):
LGTM

PR #1356, opened 2021-01-25
Fix backup-cke-etcd not working

ymmt2005:
This bug was introduced when we switched CKE installation
from container to the debian package.

`setup-serf-tags` is also fixed not to ignore the unit failure
so that we can notice the failure.


binoue (approved):
LGTM

PR #1393, opened 2021-02-08
coil: update DaemonSet Pods in bulk.

morimoto-cybozu:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,11 +40,12 @@ all:
 .PHONY: update-coil
 update-coil:
 	rm -rf /tmp/work-coil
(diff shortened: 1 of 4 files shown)


ymmt2005 (changes requested):
This is auto-generated by Makefile, and it is based on the upstream manifests.
Could you update the upstream first?

morimoto-cybozu:
@ymmt2005 
Ah, I missed that generation process.
I'll add a patch in this repository without updating the upstream.
This is because it would be appropriate to leave the update strategy to the users.

ymmt2005 (approved):
LGTM

Issue #815, opened 2020-02-26
[sabakan] Improve machine states


masa213f:
## What

Sabakan has some unused state and we want to another state, we re-consider and improve them.

- Unused state
  - `unreachable`
     There is almost no difference between `unreachable` and `unhealthy`.
     So we delete the `unreachable` and unify the abnormal states to `unhealthy`.
- New state
  - The state for healthy but not choosable because the machine is used for investigation, benchmark, or etc.

## How

- Remove `unreachable` from sabakan's machines lifecycle.
- Enable to register the reason for becoming `unhealthy`.
- Add a new state for healthy but not choosable.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

kmdkuk:
The `updating` state will be modified to be used during planned restarts. at https://github.com/cybozu-go/neco/pull/1889

PR #1472, opened 2021-03-16
leader election for sabakan-state-setter

masa213f:


pkg/sabakan-state-setter/cmd/main.go
```diff
@@ -5,37 +5,64 @@ import (
...
-	flagParallelSize   = flag.Int("parallel", 30, "parallel size")
+	flagParallelSize   = flag.Int("parallel", 30, "The number of parallel execution of getting machines metrics")
+	flagSabakanURL     = flag.String("sabakan-url", "http://localhost:10080", "sabakan URL")
+	flagSerfAddress    = flag.String("serf-address", "127.0.0.1:7373", "serf address")
 )
 
 func main() {
 	flag.Parse()
-	err := well.LogConfig{}.Apply()
+	ttl, err := time.ParseDuration(*flagEtcdSessionTTL)
```
zoetrope, line 27: FYI:
If you introduce spf13/cobra, you will be able to use `DurationVar()`.

pkg/sabakan-state-setter/controller.go
```diff
@@ -85,23 +94,111 @@ func NewController(ctx context.Context, sabakanAddress, configFile, interval str
...
+
+	election := concurrency.NewElection(session, storage.KeySabakanStateSetterLeader)
+
+	// When the etcd is stopping, the Campaign will hang up.
+	// So check the session and exit if the session is closed.
+	doneCh := make(chan error)
+	go func() {
+		doneCh <- election.Campaign(ctx, c.electionValue)
+	}()
+	select {
```
zoetrope, line 120: Don't you need to check `ctx.Done`?

pkg/sabakan-state-setter/cmd/main.go
masa213f: Thank you. `flag` pkg has `Duration()`!

https://golang.org/pkg/flag/#Duration

pkg/sabakan-state-setter/controller.go
masa213f: I will add the `ctx.Done()`.

When the `ctx` is canceled, `election.Campaign` will stop and this select statement catch an error from `doneCh`.
Therefore, there is no problem without `ctx.Done()`.
But this behavior is difficult to understand from this code.

zoetrope (commented):
I think I want a way to know who is the leader, like `ckecli leader` or REST API.
What do you think?

dctest/sabakan_state_setter_test.go
```diff
@@ -61,6 +66,76 @@ func testSabakanStateSetter() {
...
+		var leaderNodeAfter string
+		Eventually(func() error {
+			node, err := getLeaderNode(storage.KeySabakanStateSetterLeader)
+			if err != nil {
+				return err
+			}
+			leaderNodeAfter = node
+			return nil
+		}).Should(Succeed())
+		Expect(leaderNodeAfter).ShouldNot(Equal(leaderNodeBefore))
```
zoetrope, line 98: Don't you need to put this line in `Eventually`?

dctest/sabakan_state_setter_test.go
```diff
@@ -61,6 +66,77 @@ func testSabakanStateSetter() {
...
+				return errors.New("leader is not changed")
+			}
+			return nil
+		}).Should(Succeed())
+	})
+}
+
+func getLeaderNode(leaderKeyPrefix string) (string, error) {
+	stdout, _, err := execAt(bootServers[0], "env", "ETCDCTL_API=3", "etcdctl", "-w", "json",
+		"--cert=/etc/neco/etcd.crt", "--key=/etc/neco/etcd.key", "get", neco.NecoPrefix+leaderKeyPrefix, "--prefix")
```
binoue, line 105: These absolute paths for the certs are used in many place. 
How about making const vars for these?

dctest/sabakan_state_setter_test.go
```diff
@@ -61,6 +66,77 @@ func testSabakanStateSetter() {
...
+	}
+
+	revision := math.MaxInt32
+	var value string
+	for _, kvs := range result.KVS {
+		val, err := base64.StdEncoding.DecodeString(kvs.Value)
+		if err != nil {
+			return "", err
+		}
+		fmt.Printf("- %10d, %s\n", kvs.CreateRevision, string(val))
```
binoue, line 131: This is a personal preference, but I like using log.Info(..) instead of fmt.Printf(..).

dctest/sabakan_state_setter_test.go
```diff
@@ -61,6 +66,77 @@ func testSabakanStateSetter() {
...
+	}
+	err = json.Unmarshal(stdout, &result)
+	if err != nil {
+		return "", err
+	}
+	if len(result.KVS) == 0 {
+		return "", errors.New("there is no candidate")
+	}
+
+	revision := math.MaxInt32
```
binoue, line 124: Why is `math.MaxInt32` set? Var int can have a bigger value than the max value of int32 I think.

dctest/sabakan_state_setter_test.go
masa213f: That's good. I made a helper function for etcdctl.
Please check this PR. https://github.com/cybozu-go/neco/pull/1494

masa213f:
@zoetrope 
I want to know the reason of unhealthy or unreachable when I do a retiring server operation.
If we can do it, we don't need to know the leader, I think.

Let's do this task.
https://github.com/cybozu-go/neco/issues/815

zoetrope (approved):
LGTM

binoue (approved):
LGTM

PR #1493, opened 2021-03-24
Fix watchers

ysksuzuki:
This PR fixes watchers by using placemat v2. Also, updates the outdated document.


tapih (approved):
LGTM.

I confirmed that the log messages from host(boot-0) and pods are properly shown on the CircleCI dashboard.

Issue #675, opened 2019-12-08
Clear TPM and auto-shutdown retiring machines


ymmt2005:
## What

Machines whose state are `retiring` should become `retired` after its TPM
contents are cleared and shutdown.  Automate this.

## How

Extend `sabakan-state-setter` to elect a leader and let the leader responsible
to watch and change the machine states with TPM clearance and shutdown.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1495, opened 2021-03-26
Fix clients for sabakan-state-setter

masa213f:
sabakan-state-setter uses `serf.Member` and the result value of Sabakan GraphQL directly.
For that reason, the implementation is becoming more complicated in some places.

This PR changes the clients (`SabakanGQLClient` and `SerfClient`) to return only the values sabakan-state-setter needs.

And for https://github.com/cybozu-go/neco/issues/675, I change the GraphQL request to get the status and the BMC address of a machine.


binoue (approved):
LGTM

PR #1498, opened 2021-03-29
sabakan-state-setter health check

masa213f:
With https://github.com/cybozu-go/neco/issues/675, sabakan-state-setter will have two functions.

1. machine health check
    - target state: `Uninitialized`, `Healthy`, `Unhealthy` or `Unreachable`
2. machine retire
    - target state: `Retiring`

This PR renames the existing `strategy.go` to `healthcheck.go`, since the file is for the health check.
And run health check only for the machines which are `Uninitialized`, `Healthy`, `Unhealthy`, or `Unreachable` state.


binoue (approved):
LGTM

pkg/sabakan-state-setter/controller.go
```diff
@@ -199,27 +203,116 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+		}
+	}
+	for undefinedType, invalidMachines := range undefinedMachineTypes {
+		if undefinedType == "" {
+			log.Warn("machine type is not specified", map[string]interface{}{
+				"machines": strings.Join(invalidMachines, ","),
+			})
+		} else {
+			log.Warn("specified machine type does not exist", map[string]interface{}{
+				"machine-type": undefinedType,
```
morimoto-cybozu, line 228: A hyphen cannot be used for field names.
https://github.com/cybozu-go/log/blob/v1.6.0/formatter.go#L42

pkg/sabakan-state-setter/controller.go
```diff
@@ -199,27 +203,116 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+	// Do machines health check
+	newStateMap := c.machineHealthCheck(ctx, machines)
+
+	// Do machines retire
+	// T.B.D.
+
+	now := time.Now()
+	for _, m := range machines {
+		newState, ok := newStateMap[m.Serial]
+		if !ok || m.State == newState {
```
morimoto-cybozu, line 243: The check of `m.State == newState` is new in this PR.
It seems reasonable, but I'm afraid there is some reason not to do this check in the old code...

By the way, `newState == m.State` would be preferable.

pkg/sabakan-state-setter/controller.go
```diff
@@ -199,27 +203,116 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+			continue
+		}
+
+		if newState == sabakan.StateUnhealthy {
+			if ok := c.RegisterUnhealthy(m, now); !ok {
+				continue
+			}
+		} else {
+			c.ClearUnhealthy(m)
```
morimoto-cybozu, line 252: This is not called if a machine transits from Unhealthy to Retiring because a retiring machine is not included in the newStateMap.
It leaves an entry in `c.unhealthyMachines`, while it affects just a little.

pkg/sabakan-state-setter/controller.go
```diff
@@ -255,51 +347,18 @@ func (c *Controller) runOnce(ctx context.Context) error {
 	}
 	wg.Wait()
 
-	now := time.Now()
-
+	// Decide next machine state
+	newStateMap := map[string]sabakan.MachineState{}
 	for _, mss := range machineStateSources {
+		log.Info("do health check", map[string]interface{}{
```
morimoto-cybozu, line 353: Do we need this log?

pkg/sabakan-state-setter/controller.go
```diff
@@ -255,51 +347,18 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+			"serial": mss.serial,
+			"ipv4":   mss.ipv4,
+		})
 		newState, hasTransition := mss.decideMachineStateCandidate()
```
morimoto-cybozu, line 357: `hasTransition` seems misleading.

pkg/sabakan-state-setter/controller_test.go
```diff
@@ -8,21 +8,34 @@ import (
...
-		machineTypes:      map[string]*machineType{mt.Name: mt},
+		machineTypes:      machineTypes,
 		unhealthyMachines: make(map[string]time.Time),
 	}
 }
 
 func testControllerRun(t *testing.T) {
 	t.Parallel()
 
+	machineTypeSkip := &machineType{
```
morimoto-cybozu, line 31: `machineTypeNoMetricsCheck` or `machineTypeSerfOnly` would be more meaningful.

pkg/sabakan-state-setter/controller_test.go
```diff
@@ -60,50 +73,87 @@ func testControllerRun(t *testing.T) {
...
-	gql = newMockGQLClient()
-	serf, _ = newMockSerfClient()
-	metricsInput = `
+`},
+			expected: map[string]sabakan.MachineState{
+				"00000001": sabakan.StateUnhealthy,
+			},
+		},
+		{
+			message: "healthcheck: transition machine state to unhealthy due to warning disks become larger than one",
```
morimoto-cybozu, line 111: The test messages need refining.

morimoto-cybozu (changes requested):

pkg/sabakan-state-setter/controller.go
masa213f: I am one of the members who wrote the initial sabakan-state-setter.
As far as I remember, there is not any reason in the old code...

masa213f: Without the `!ok` check, the state of the StateUpdating, StateRetiring, or StateRetired machines will be updated unnecessarily.
So this condition needs `!ok`.

pkg/sabakan-state-setter/controller_test.go
masa213f: Update the tests. de6a72b
The test needs to do the health checks of multiple machines. So I have updated this test with the message.

pkg/sabakan-state-setter/controller.go
masa213f: I changed the location of `ClearUnhealthy`.
In most cases, this leak will be prevented by this fix.
9468abd

pkg/sabakan-state-setter/controller.go
morimoto-cybozu: Checking `!ok` is OK:-)
I have a feeling of strangeness in the order of the equality check, though it might be a matter of taste.

```suggestion
		if !ok || newState == m.State {
```

pkg/sabakan-state-setter/controller.go
```diff
@@ -199,27 +202,105 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+		if !ok || m.State == newState {
+			c.ClearUnhealthy(m)
+			continue
+		}
+		if newState == sabakan.StateUnhealthy {
+			// Wait for the GracePeriod before changing the machine state to unhealthy.
+			if ok := c.RegisterUnhealthy(m, now); !ok {
+				continue
+			}
+		}
```
morimoto-cybozu, line 241: It would be good to call `c.ClearUnhealthy(m)` in an `else` clause.
Even without that call, `c.unhealthyMachines[m.Serial]` will be cleared in the next invocation of `runOnce()` because `newState == m.State` will become true in most cases.
But I don't think it is intuitive.

Using `switch` might help.

```
switch {
case !ok || newState == m.State:
    c.CleanUnhealthy(m)
    continue
case newState == sabakan.Stateunhealthy:
    if ok := c.RegisterUnhealthy(m, now); !ok {
        continue
    }
default:
    c.CleanUnhealthy(m)
}
```

pkg/sabakan-state-setter/controller.go
```diff
@@ -199,27 +202,105 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+func (c *Controller) machineHealthCheck(ctx context.Context, machines []*machine) map[string]sabakan.MachineState {
+	// Get serf status
 	serfStatus, err := c.serfClient.GetSerfStatus()
 	if err != nil {
 		log.Warn("failed to get serf members", map[string]interface{}{
 			log.FnError: err.Error(),
 		})
-		// lint:ignore nilerr  RunPeriodically tries this again.
-		return nil
+		return map[string]sabakan.MachineState{}
```
morimoto-cybozu, line 283: The returned map will not be modified, so we don't need to make an empty map here.

```suggestion
		return nil
```

morimoto-cybozu (changes requested):

pkg/sabakan-state-setter/controller_test.go
```diff
@@ -112,63 +151,173 @@ hw_storage_controller_status_health{controller="PCIeSSD.Slot.2-C", system="Syste
...
-		if err != nil {
-			t.Error(err)
-		}
-		time.Sleep(100 * time.Millisecond)
+`},
+			expected: map[string]sabakan.MachineState{
+				"00000000": sabakan.StateUnhealthy,   // cpu is warning
+				"00000001": sabakan.StateUnhealthy,   // warning disks become larger than one
+				"00000002": sabakan.StateHealthy,     // warning disks is one
+				"00000003": sabakan.StateUnreachable, // serf status is failed
```
morimoto-cybozu, line 159:
```suggestion
				"00000000": sabakan.StateUnhealthy,   // one of two CPUs is issuing a warning
				"00000001": sabakan.StateUnhealthy,   // all HDD are unhealthy; # of healthy HDDs falls below MinimumHealthyCount (0 < 1)
				"00000002": sabakan.StateHealthy,     // one of two HDDs is unhealthy, but it is acceptable
				"00000003": sabakan.StateUnreachable, // serf status is "failed"
```

masa213f:
@morimoto-cybozu 
I reflected comments and check the behavior on the dctest.
Please review again.

morimoto-cybozu (approved):
LGTM

PR #1510, opened 2021-04-06
sabakan-state-setter: Exit if getting serf status is failed

masa213f:
If the serf service is restarted, the connection of the serf client is closed.
In this case, `GetSerfStatus()` will never succeed unless recreating the client.

So return an error to exit sabakan-state-setter.


binoue (approved):
LGTM

PR #1524, opened 2021-04-12
neco reboot-worker: skip reboot if machine power is already off

masa213f:
`neco reboot-worker` will fail if there are machines powered off.

### DIFF
--- a/pkg/neco/cmd/reboot_worker.go
+++ b/pkg/neco/cmd/reboot_worker.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cybozu-go/neco"
 	"github.com/cybozu-go/well"
 	"github.com/spf13/cobra"
+	"github.com/stmcginnis/gofish/redfish"
 	"github.com/tcnksm/go-input"
 	"sigs.k8s.io/yaml"
 )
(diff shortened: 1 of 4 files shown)


pkg/neco/cmd/reboot_worker.go
```diff
@@ -192,7 +192,11 @@ func rebootMain() error {
 		})
 		err := power(context.Background(), "restart", addr)
 		if err != nil {
-			return err
+			log.Warn("failed to restart machine", map[string]interface{}{
+				"serial":    m.Spec.Serial,
+				"node":      m.Spec.IPv4[0],
+				log.FnError: err,
+			})
```
morimoto-cybozu, line 199: By ignoring errors, the whole command exits successfully even when it has failed to restart some machine.
The importance of the failure varies according to the cause of the failure and the intention of the user, so it would be nice to notify the user of the failure via the exit code.
Then the user will be able to judge what to do.

How about remembering the last error and returning it finally.
It is not necessary to keep a slice of errors because the user can investigate the errors from the warning messages.

masa213f: You're right. I fixed it.

morimoto-cybozu (approved):
LGTM

PR #1532, opened 2021-04-14
sabakan-state-setter: shutdown retired machines

masa213f:
Add a new feature to `sabakan-state-setter` to auto shutdown retired machines.
When this change is released, the retired machines will be shut down at 1:00 AM (JST) every day.

### DIFF
--- a/etc/sabakan-state-setter.yml
+++ b/etc/sabakan-state-setter.yml
@@ -1,4 +1,4 @@
-shutdown-schedule: 00 16 * * *
+shutdown-schedule: 0 16 * * *
 machine-types:
   - name: qemu
     grace-period: 5s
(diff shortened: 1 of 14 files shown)


ymmt2005 (commented):
LGTM

etc/sabakan-state-setter.yml
```diff
@@ -1,3 +1,4 @@
+shutdown-schedule: 00 16 * * *
```
morimoto-cybozu, line 1: Just a piece of trivia:-)
"00" is not common for crontab.  This type of notation sometimes causes errors when used with non-sophisticated libraries.  Especially "08" and "09" can be wrongly considered as illegal octal numbers.
Go's `strconv.Atoi()` is OK.  `robfig/cron` uses `Atoi()` and it's OK too.

masa213f: Thank you. I fixed it just in case.

binoue (approved):
mobbed

PR #1533, opened 2021-04-15
Add safety features to 'neco tpm clear'

masa213f:
- Add safety features to `neco tpm clear`. (same as [`sabactl crypts delete`](https://github.com/cybozu-go/sabakan/blob/main/docs/sabactl.md#sabactl-crypts-delete-serial))
    - The command fails when the target machine's status is not `Retiring`.
    - The command fails when `--force` option is not specified.
- Output failure messages of `neco tpm clear` in the sabakan-state-setter log.

### DIFF
--- a/pkg/sabakan-state-setter/necocommand.go
+++ b/pkg/sabakan-state-setter/necocommand.go
@@ -1,14 +1,13 @@
 package sss
 
 import (
-	"bytes"
 	"context"
 	"os/exec"
 )
 
 // NecoCmdExecutor is interface for the neco command
 type NecoCmdExecutor interface {
-	TPMClear(ctx context.Context, serial string) (string, error)
+	TPMClear(ctx context.Context, serial string) ([]byte, error)
 }
 
 type necoCmdExecutor struct{}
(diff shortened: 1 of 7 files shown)


pkg/sabakan-state-setter/necocommand.go
```diff
@@ -17,15 +17,17 @@ func newNecoCmdExecutor() necoCmdExecutor {
...
-func execNeco(ctx context.Context, opts ...string) (string, error) {
+func execNeco(ctx context.Context, opts ...string) (string, string, error) {
 	cmd := exec.CommandContext(ctx, "neco", opts...)
 	outBuf := new(bytes.Buffer)
+	errBuf := new(bytes.Buffer)
 	cmd.Stdout = outBuf
+	cmd.Stderr = errBuf
 	err := cmd.Run()
-	return outBuf.String(), err
+	return outBuf.String(), errBuf.String(), err
```
ymmt2005, line 27: Why don't you use `CombinedOutput()`?

ymmt2005 (changes requested):
LGTM

pkg/sabakan-state-setter/necocommand.go
masa213f: That's good. I fixed it.

ymmt2005 (approved):
LGTM

Issue #1538, opened 2021-04-16
Refine management of pstore


morimoto-cybozu:
## What

Systemd introduced a new service `systemd-pstore.service` in v243.
https://github.com/systemd/systemd/blob/v243/NEWS#L378
https://www.phoronix.com/scan.php?page=news_item&px=Systemd-Pstore-Service
https://www.freedesktop.org/software/systemd/man/systemd-pstore.service.html

This services moves files from `/sys/fs/pstore` to `/var/lib/systemd/pstore` in default.
In our environment, this means that files are transferred from small-but-persistent storage to large-but-volatile memory.

Refine the management of pstore not to lose the contents with successive reboot while managing the limited space of pstore.

## How

Mount logical volume to `/var/lib/systemd`. 

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #1541, opened 2021-04-19
make logical volume to persist /var/lib/systemd

yamatcha:
Implement #1538


ymmt2005 (changes requested):
Please update the document too.
https://github.com/cybozu-go/neco/blob/main/docs/ignition.md#mount-lvm-volumes-under-var

ymmt2005 (approved):
LGTM

Issue #1543, opened 2021-04-19
[neco] Prevent `neco-updater` from installing older neco release versions


binoue:
## What

GitHub may accidentally return an older release version, as happened in [this incident](https://bozuman.cybozu.com/k/#/space/3801/thread/29612/2849682).
To prevent this, we need you to modify `neco-updater`.

## How

When `neco-updater` detects an older release version, output an error and ignore it.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1547, opened 2021-04-20
neco-worker: Update Neco only when the package is newer than existing

tapih:
Implement #1543

### DIFF
--- a/updater/release_checker.go
+++ b/updater/release_checker.go
@@ -90,7 +90,7 @@ func (c *ReleaseChecker) update(ctx context.Context) error {
 		return err
 	}
 
-	if c.current == latest {
+	if latest == c.current {
 		return nil
 	}
 
--- a/updater/release_checker.go
+++ b/updater/release_checker.go
@@ -3,10 +3,7 @@ package updater
 import (
 	"context"
 	"errors"
-	"fmt"
 	"net/http"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/cybozu-go/log"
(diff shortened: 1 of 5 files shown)


worker/worker.go
```diff
@@ -87,7 +87,7 @@ func (w *Worker) Run(ctx context.Context) error {
 			continue
 		}
 
-		if w.version != req.Version {
```
ymmt2005, line 90: Check the version in `neco-updater` rather than in `neco-worker`.
Or you would get a timeout error from `neco-updater`.

ymmt2005 (changes requested):

updater/release_checker.go
```diff
@@ -90,7 +93,28 @@ func (c *ReleaseChecker) update(ctx context.Context) error {
 		return err
 	}
 
-	if latest == c.current {
+	if c.current == latest {
```
ymmt2005, line 96: I think a local value should come first for binary operators:
```suggestion
	if latest == c.current {
```

updater/release_checker.go
```diff
@@ -101,3 +125,29 @@ func (c *ReleaseChecker) update(ctx context.Context) error {
...
+		return nil, err
+	}
+	v, err := strconv.Atoi(tags[2])
+	if err != nil {
+		return nil, err
+	}
+	return &necoRelease{tags[0], d, v}, nil
+}
+
+func (r necoRelease) isNewerThan(target *necoRelease) (bool, error) {
```
ymmt2005, line 151: Could you please add tests?

ymmt2005 (changes requested):

ymmt2005 (approved):
LGTM

PR #1554, opened 2021-04-22
ignitions: disable kernel.hung_task_panic

ymmt2005:
Programs that use Ceph RBD block device can hang longer than
120 seconds, and let the kernel panic because of this setting.

Disable it to prevent panics at least for now.

PR #1555, opened 2021-04-22
Update coil to 2.0.7

ysksuzuki:


ysksuzuki:
#1554 will update coil

PR #1560, opened 2021-04-26
Fix release version check to use proper format

tapih:
Current version check is incomplete as the version check had a bug and was reverted.
This PR fix the bug.  

Signed-off-by: Hiroshi Muraoka <h.muraoka714@gmail.com>


binoue (approved):
LGTM

Issue #1586, opened 2021-05-14
add options to `neco reboot-workers" to specify the workers to be rebooted


umezawatakeshi:
## What

add options to `neco reboot-workers" to specify the workers to be rebooted.
it makes per-rack reboot possible, for example.

## How

add options and get workers by `sabactl machines get` or its equivalent.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

Issue #1577, opened 2021-05-10
Apply individual firmware updaters and reboot automatically


umezawatakeshi:
## What

automate individual updater EXE based firmware update procedure

## How

- add firmware application (upload) feature to setup-hw
- add automation/reboot logic to neco
- write document

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1587, opened 2021-05-14
add `sabactl machines get`-like options to neco reboot-workers.

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1586


docs/neco.md
```diff
@@ -185,11 +185,12 @@ Synopsis
 
 * `neco reboot-worker`
 
-    Reboot all worker nodes.
+    Reboot all or specified worker nodes.
 
     This uses CKE's function of [graceful reboot](https://github.com/cybozu-go/cke/blob/main/docs/reboot.md) for the nodes used by CKE.
     As for the other nodes, this reboots them immediately.
     If some nodes are already powered off, this command does not do anything to those nodes.
+    `sabactl machines get`-like options can be used to narrow down the machines to be rebooted.
```
yokaze, line 193:
```suggestion
    Options like `sabactl machines get` can be used to narrow down the machines to be rebooted.
```
I prefer to write "Options like `sabactl machines get`", but I leave it to your choice.

pkg/neco/cmd/reboot_worker.go
```diff
@@ -130,6 +105,10 @@ func rebootMain() error {
 
 	var ss, nonss []string
 	for _, node := range cluster.Nodes {
+		if _, ok := machineAddrs[node.Address]; !ok {
+			continue
+		}
+
```
yokaze, line 111: I could not find there is any chance of invalid addresses, since they are retrieved using `sabakanMachinesGetExcludingBootServers`. It looks redundant to me.

pkg/neco/cmd/reboot_worker.go
```diff
@@ -107,15 +69,28 @@ If some nodes are already powered off, this command does not do anything to thos
...
 
+	return rebootMachines(machines)
+}
+
+func rebootMachines(machines []sabakan.Machine) error {
+	machineAddrs := make(map[string]struct{}, len(machines))
+	for _, m := range machines {
+		if m.Spec.Role == "boot" {
+			return fmt.Errorf("it's not allowed to reboot boot servers")
+		}
```
yokaze, line 90: It looks like an unneeded double check since the machines are retrieved using `sabakanMachinesGetExcludingBootServers`.
I prefer to write a single comment that indicates the assumption on the function argument.

pkg/neco/cmd/sabakan_machines_get.go
```diff
@@ -0,0 +1,64 @@
...
+	params := make(map[string]string)
+	for k, v := range opts.params {
+		params[k] = *v
+	}
+	c, err := client.NewClient(neco.SabakanLocalEndpoint, httpClient.Client)
+	if err != nil {
+		return nil, err
+	}
+	return c.MachinesGet(ctx, params)
+}
```
yokaze, line 48: It looks redundant to me because we can just use `sabactl machines get`.
A good reason should be provided if it is important.

pkg/neco/cmd/sabakan_machines_get.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	ret := []sabakan.Machine{}
+	for _, m := range machines {
+		if m.Spec.Role != "boot" {
+			ret = append(ret, m)
+		}
+	}
+	return ret, nil
+}
```
yokaze, line 64: Without `sabakanMachinesGet`, it can be written in `reboot_worker.go`.

yokaze (changes requested):

pkg/neco/cmd/reboot_worker.go
umezawatakeshi: By adding selection options, there are some machines which are returned by `ckecli cluster get` but are not returned by `sabactl machines get`. Those machines should not be rebooted.

pkg/neco/cmd/reboot_worker.go
umezawatakeshi: This function is to be called by other part. (see https://github.com/cybozu-go/neco/issues/1577)
So I added this check by beforehand.

docs/neco.md
morimoto-cybozu: A link for the `sabactl` document should be here.

pkg/neco/cmd/reboot_worker.go
```diff
@@ -107,15 +69,28 @@ If some nodes are already powered off, this command does not do anything to thos
...
+	machines, err := sabakanMachinesGetExcludingBootServers(ctx, &rebootWorkerGetOpts)
 	if err != nil {
 		return err
 	}
 
+	return rebootMachines(machines)
+}
+
+func rebootMachines(machines []sabakan.Machine) error {
+	machineAddrs := make(map[string]struct{}, len(machines))
```
morimoto-cybozu, line 86: I prefer `map[string]bool` for readability.
https://stackoverflow.com/questions/37320287/maptstruct-and-maptbool-in-golang

pkg/neco/cmd/sabakan_machines_get.go
umezawatakeshi: most part of neco use sabakan client object.

pkg/neco/cmd/sabakan_machines_get.go
yokaze: I'm not sure to have a file here that does not correspond to a command.
How do you think @morimoto-cybozu -san? I would like to have another opinion.

yokaze: s/I'm not sure/I'm not sure it is good/g

pkg/neco/cmd/sabakan_machines_get.go
yokaze: I thought the function is for user-facing command, but it is not. Sorry for my misunderstanding.

pkg/neco/cmd/reboot_worker.go
yokaze: I understand.

pkg/neco/cmd/sabakan_machines_get.go
morimoto-cybozu: I prefer to include these function in `reboot_worker.go`.
The existence of `sabakan_machines_get.go` makes me suppose that `neco sabakan machines get` command is available.

yokaze: @umezawatakeshi 
I suspect the function is in a separate file (`sabakan_machines_get.go`) because it will be used in another command.
I think there are two options for the function:
- write it in `reboot_worker.go` and use it from multiple commands, or
- rename it as `opts_sabakan_machines_get.go` or some other name that indicates it is not a command.

Or if the function is used only from `reboot_worker.go`, I prefer it written in `reboot_worker.go`.
Can you please consider a change?

pkg/neco/cmd/reboot_worker.go
```diff
@@ -130,6 +112,10 @@ func rebootMain() error {
 
 	var ss, nonss []string
 	for _, node := range cluster.Nodes {
+		if machineAddrs[node.Address] {
```
morimoto-cybozu, line 115:
```suggestion
		if !machineAddrs[node.Address] {
```

morimoto-cybozu (approved):
LGTM

yokaze (approved):
LGTM

PR #1591, opened 2021-05-17
Remove test which should be deleted after #1560 updated

yamatcha:
Related #1560 

Signed-off-by: yamatcha <soju-yamashita@cybozu.co.jp>


binoue (approved):
LGTM

PR #1617, opened 2021-05-25
Update etcd clients to use v3.4.16 API

yokaze:
This PR make neco CLI tools to use etcd 3.4 APIs.
`etcdctl` is not intended to be upgraded in this PR.


ymmt2005 (approved):
LGTM

PR #1623, opened 2021-05-27
Record all k8s events in the audit log

yokaze:
This PR records all the k8s events in the audit log.
They are recorded using `create` and `patch` verbs; read and delete operations of the events are not recorded.


ymmt2005 (approved):
LGTM

Issue #1647, opened 2021-06-09
limit local ephemeral storage, again


zoetrope:
## What

In order to specify the limit for local ephemeral storage, `/var/log` and `/var/lib/kubelet` must be in a same filesystem.

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage

## How

- Use bind mount in fstab in ignition.
- Revert https://github.com/cybozu/neco-containers/pull/623

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1650, opened 2021-06-10
bind mount /var/log to /var/lib/kubelet/varlog

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1647

- bind mount
- always re-mkfs /var/lib/kubelet to make it "volatile"


yokaze (approved):
LGTM

PR #1654, opened 2021-06-15
Use raw image for boot servers

ysksuzuki:


ysksuzuki:
Converting to raw format did not work.

PR #1676, opened 2021-06-29
[reboot-and-wait] Wait for machine to become healthy

morimoto-cybozu:


binoue (approved):
LGTM. 

It might be good to update [the description of `neco reboot-and-wait` command](https://github.com/cybozu-go/neco/blob/main/docs/neco.md#bmc-management-functions), but I do not have a strong opinion about this.

PR #1678, opened 2021-06-30
Revert "add kubectl-hns to tools"

binoue:
We will use [Innu](https://github.com/cybozu-go/innu) in future, so remove "kubectl-hns" .


morimoto-cybozu (approved):
LGTM

Issue #1677, opened 2021-06-30
core dump handling workflow


umezawatakeshi:
## What

- implement workflow to manage Pod's core dump
- add documentation

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #1679, opened 2021-06-30
add dedicated LV for /var/lib/systemd/coredump

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1677

### DIFF
--- a/ignitions/common/files/opt/sbin/setup-var
+++ b/ignitions/common/files/opt/sbin/setup-var
@@ -1,11 +1,14 @@
 #!/bin/sh -eu
 
 VG=vg1
-LVLIST="k8s-containerd docker kubelet systemd"
+LVLIST="k8s-containerd docker kubelet systemd coredump"
+
+# these values are for qemu (dctest env). see below for real machines.
 CONTAINERD_SIZE=20g
 DOCKER_SIZE=10g
 KUBELET_SIZE=10g
 SYSTEMD_SIZE=1g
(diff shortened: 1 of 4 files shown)


ignitions/common/files/opt/sbin/setup-var
```diff
@@ -81,12 +87,14 @@ elif ls /dev/mapper/crypt-nvme* >/dev/null 2>&1; then
     CONTAINERD_SIZE=1t
     DOCKER_SIZE=50g
     KUBELET_SIZE=100g
+    COREDUMP_SIZE=50g
```
ysksuzuki, line 90: Why is `COREDUMP_SIZE` 50G for cs and 20G for ss? I read the argument that 10G or more is enough though.

ysksuzuki:
JVM crash dumps suppose to be stored `/var/lib/systemd/coredump` for Java applications such as Elasticsearch?

ysksuzuki:
> JVM crash dumps suppose to be stored /var/lib/systemd/coredump for Java applications such as Elasticsearch?

Sorry , forget it. JVM crush dump is not in our scope.

ignitions/common/files/opt/sbin/setup-var
umezawatakeshi: the partition should be larger enough to hold uncompressed core.

ysksuzuki (approved):

PR #1692, opened 2021-07-07
fix kernel.core_pattern in sysctl

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1677

This PR solves the problem that systemd-coredump does not retrieve the correct process information.

Use the default value of kernel.core_pattern, which is set by /usr/lib/sysctl.d/50-coredump.conf


yokaze:
@umezawatakeshi What problem this PR solves?

umezawatakeshi:
This PR solves the problem that systemd-coredump does not retrieve the correct process information.

ysksuzuki (approved):

Issue #1694, opened 2021-07-09
using perf on pod


umezawatakeshi:
## What

Establish how to use perf  on pod by tenant users

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #1701, opened 2021-07-19
change settings for non-privilege users to use perf

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1694

- set `kernel.perf_event_paranoid` sysctl to `1`
- ~remount~ mount /sys/kernel/tracing with mode=755

### DIFF
--- a/ignitions/common/common.yml
+++ b/ignitions/common/common.yml
@@ -17,6 +17,7 @@ files:
   - /etc/systemd/coredump.conf
   - /etc/systemd/system.conf.d/50-cybozu.conf
   - /etc/systemd/system/rngd.service.d/non_vm.conf
+  - /etc/systemd/system/sys-kernel-tracing.mount.d/mode.conf
   - /etc/systemd/system/systemd-journald.service.d/oom_score_adj.conf
   - /opt/bin/load-containerd-image
   - /opt/bin/load-docker-image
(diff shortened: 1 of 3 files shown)


ignitions/common/systemd/remount-sys-kernel-tracing.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=remount /sys/kernel/tracing
+After=network.target
+# "After=network.target" is same as rc-local.service
+
+[Service]
+Type=oneshot
+ExecStart=mount -o remount,mode=755 /sys/kernel/tracing/
```
morimoto-cybozu, line 8: @umezawatakeshi 
Have you tried a drop-in file `ignitions/common/files/etc/systemd/system/sys-kernel-tracing.mount.d/mode.conf` to change the mount options directly?

```
[Mount]
Options=mode=755
#Options=nosuid,nodev,noexec,mode=755  # if "Options" is not cumulative; in this case we need to chase the changes in the upstream, which is not desirable
```

umezawatakeshi: @morimoto-cybozu 
No, I haven't. I will try it.
How/where did you find out this feature?

morimoto-cybozu: @umezawatakeshi 
https://www.freedesktop.org/software/systemd/man/systemd.unit.html
> Along with a unit file foo.service, a "drop-in" directory foo.service.d/ may exist. All files with the suffix ".conf" from this directory will be merged in the alphanumeric order and parsed after the main unit file itself has been parsed.

The description above uses "foo.service" for example, but the drop-in directory would be common functionality for other unit files including "foo.mount."
I have not tested yet, though:-)

umezawatakeshi: succeeded on my local environment

morimoto-cybozu (approved):
LGTM

PR #1724, opened 2021-08-17
use HWE kernel for boot server installer

umezawatakeshi:
We need a driver for Intel E810-XXV.


ymmt2005 (approved):
LGTM

Issue #1727, opened 2021-08-18
the actual behavior of `neco recover` is different from its documentation.


umezawatakeshi:
**Describe the bug**

[the documentation](https://github.com/cybozu-go/neco/blob/e8c91f7fbfb2dacf7739dcb4417e19285664079c/docs/update.md#sabakan-contents-eg-container-images-os-images-and-ignitions) says that `neco recover` remove `<prefix>/contents/sabakan` key in etcd, but it does not do so actually.

I'm not sure which (or both) should be fixed.

**Environments**
- Version:

PR #1729, opened 2021-08-19
follow `neco recover` behavior to current neco-worker process.

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1727


morimoto-cybozu (approved):
paired

Issue #1738, opened 2021-09-02
Check status code and retry the deb package download


kmdkuk:
## What

- If GitHub returns a 50X error when downloading a deb package, retry the download several times.

## How

- `Client.Get` in `net/http` package succeeds only if the status code is in the 20x series, and to redirect.
  - https://pkg.go.dev/net/http#Client.Get
- If the status code is in the 50x series, it needs to be handled by retrying the request.
- The target code is around here.
  - https://github.com/cybozu-go/neco/blob/1050fd9666e6d7d5dbe44f665341a73b60cdd6c4/worker/deb.go#L22-L31

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1740, opened 2021-09-03
check status code and retry the deb package download

kmdkuk:
#1738 

There is no clear intention for the 10 of retry count and 10s of retry intervals.

### DIFF
--- a/worker/deb.go
+++ b/worker/deb.go
@@ -7,7 +7,9 @@ import (
 	"net/http"
 	"os"
 	"regexp"
+	"time"
 
+	"github.com/cybozu-go/log"
 	"github.com/cybozu-go/neco"
 	"github.com/cybozu-go/well"
 	"github.com/google/go-github/v35/github"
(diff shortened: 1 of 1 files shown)


worker/deb.go
```diff
@@ -24,19 +27,42 @@ func InstallDebianPackage(ctx context.Context, client *http.Client, ghClient *ht
...
+			}
+			defer resp.Body.Close()
+			if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
+				return fmt.Errorf("status code %d is not a success", resp.StatusCode)
+			}
+			_, err = io.Copy(debPkg, resp.Body)
+			if err != nil {
+				return err
+			}
+			return nil
```
ymmt2005, line 46:
```suggestion
			return err
```

worker/deb.go
```diff
@@ -24,19 +27,42 @@ func InstallDebianPackage(ctx context.Context, client *http.Client, ghClient *ht
...
+		func(ctx context.Context) error {
+			resp, err := client.Get(downloadURL)
+			if err != nil {
+				return err
+			}
+			defer resp.Body.Close()
+			if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
+				return fmt.Errorf("status code %d is not a success", resp.StatusCode)
+			}
+			_, err = io.Copy(debPkg, resp.Body)
```
ymmt2005, line 42: What happens if `io.Copy` fails (due to network error or something) then this function is retried?

I'm worrying about there would be malformed data after/before the complete deb data.

ymmt2005 (changes requested):

worker/deb.go
kmdkuk: thank you! fix it.

worker/deb.go
kmdkuk: I think you are right in your worrying.
I fixed it to do `Reset()` before `io.Copy`.

ymmt2005: Or you can use `io.ReadAll()` and store the result data, like this.

```go
var data []byte
... (snip)
    data, err = io.ReadAll(resp.Body)
    if err != nil { return err }
```

kmdkuk: That's going to be simpler.

ymmt2005 (approved):
LGTM

Issue #1717, opened 2021-08-06
Add amtool to teleport node pods


yokaze:
## What

Describe what this issue should address.

## How

Describe how to address the issue.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

ymmt2005:
@yokaze renamed the title so that it describes what should be done properly.

PR #1742, opened 2021-09-06
add amtool to operation cli packages

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1717


yokaze (approved):
LGTM

PR #1743, opened 2021-09-06
make systemd-pstore run after mounting pstore

umezawatakeshi:
In some environment, systemd-pstore.service run **before** /sys/fs/pstore is mounted. I doesn't make sense.


ymmt2005 (approved):
LGTM

Issue #1803, opened 2021-10-28
Improve power management of retired machines


morimoto-cybozu:
### Problems

A Dell server machine equipped with TPM 1.2 failed to be powered-off by sabakan-state-setter when its state transited from "retiring" to "retired."
We couldn't get detailed logs from sabakan-state-setter.
We couldn't notice the zombie machine until the metrics of the API servers complained about too many errors.
sabakan-state-setter reported "shutdown successfully" about periodic power-off of retired machines, but a machine continued to run indeed.

### Causes

sabakan-state-setter relies on "neco tpm clear" for changing the power state of a retiring machine.
https://github.com/cybozu-go/neco/blob/main/docs/sabakan-state-setter.md
"neco tpm clear" does nothing for TPM 1.2 machines though it should change the power state at least.

sabakan-state-setter does not record the outputs from "neco tpm clear."

We don't configure an alert about zombie machines, i.e. retired but living machines.

## How

Describe how to address the issue.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1807, opened 2021-11-01
Improve power management of retired/retiring machines

masa213f:
https://github.com/cybozu-go/neco/issues/1803

- Use `ForceOff` in `neco power stop`.
- Power off if clearing TPM is not supported in `neco tpm clear`.
- Output the result of `neco` command from `sabakan-state-setter`.

### DIFF
--- a/pkg/neco/cmd/tpm_clear.go
+++ b/pkg/neco/cmd/tpm_clear.go
@@ -164,12 +164,14 @@ IP is one of the IP addresses owned by the machine.`,
 			machineType := machine.Spec.Labels[machineTypeLabelName]
 			logicType, ok := supportedMachineTypes[machineType]
 			if !ok {
+				err := power(ctx, "stop", machine.Spec.BMC.IPv4)
 				log.Warn("unknown machine type; shutdown", map[string]interface{}{
-					"serial":       machine.Spec.Serial,
-					"node":         machine.Spec.IPv4[0],
-					"machine_type": machineType,
+					"serial":       "machine.Spec.Serial",
+					"node":         "machine.Spec.IPv4[0]",
+					"machine_type": "machineType",
+					"result":       err,
 				})
-				return power(ctx, "stop", machine.Spec.BMC.IPv4)
+				return errors.New("unknown machine type")
 			}
 
 			switch logicType {
@@ -178,7 +180,7 @@ IP is one of the IP addresses owned by the machine.`,
 			case tpmClearLogicTypeDellRedfish:
 				return dellRedfishClearTpm(ctx, machine.Spec.BMC.IPv4)
 			default:
-				panic(fmt.Sprintf("unknown logic type: %d", logicType))
+				panic(fmt.Sprintf("unreachable; unknown logic type: %d", logicType))
 			}
 		})
 		well.Stop()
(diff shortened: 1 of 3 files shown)


pkg/neco/cmd/tpm_clear.go
```diff
@@ -165,19 +164,21 @@ IP is one of the IP addresses owned by the machine.`,
...
-			case tpmClearLogicTypeNothing:
-				log.Info("nothing to do", nil)
-				return nil
+			case tpmClearLogicTypePowerOff:
+				return power(ctx, "stop", machine.Spec.BMC.IPv4)
 			case tpmClearLogicTypeDellRedfish:
 				return dellRedfishClearTpm(ctx, machine.Spec.BMC.IPv4)
 			default:
-				return fmt.Errorf("unknown logic type: %d", logicType)
+				panic(fmt.Sprintf("unknown logic type: %d", logicType))
```
morimoto-cybozu, line 181: I'm not sure why returning an error is inappropriate.

pkg/neco/cmd/tpm_clear.go
```diff
@@ -165,19 +164,21 @@ IP is one of the IP addresses owned by the machine.`,
 			machineType := machine.Spec.Labels[machineTypeLabelName]
 			logicType, ok := supportedMachineTypes[machineType]
 			if !ok {
-				return fmt.Errorf("unknown machine type: machine-type=%s", machineType)
+				log.Warn("unknown machine type; shutdown", map[string]interface{}{
+					"serial":       machine.Spec.Serial,
+					"node":         machine.Spec.IPv4[0],
+					"machine_type": machineType,
+				})
+				return power(ctx, "stop", machine.Spec.BMC.IPv4)
```
morimoto-cybozu, line 172: To power-off anyway would be good.

I prefer to return an error here even when power-off succeeds, because the operator expects "neco tpm clear" both to clear the TPM devices and to power-off the machines.
In this case the TPM devices are not cleared, so not all expectations are fulfilled.
This situation should be expressed by an error rather than a warning.

For QEMU and TPM 1.2 machines, the operator should not expect the command to clear the TPM devices, so returning non-error would be appropriate.

morimoto-cybozu (changes requested):

pkg/neco/cmd/tpm_clear.go
masa213f: That's right. Functionally both returning an error and calling `panic()` are almost the same.

But this switch statement only accepts the following constants. And this `default` case is not reached.
So I thought it would be suitable to use `panic()`.

```
const (
	tpmClearLogicTypePowerOff = iota // If clearing TPM is not supported, shutdown the machine.
	tpmClearLogicTypeDellRedfish
)
```

pkg/neco/cmd/tpm_clear.go
masa213f: That's good. https://github.com/cybozu-go/neco/pull/1807/commits/212012ef740c764f494fd96cafd0853fd60402b3
When logicType for a target machine is not defined, `neco tpm clear` will output the following logs.

```
2021-11-02T02:19:03.222572Z ubuntu-local neco warning: "unknown machine type; shutdown" machine_type="MACHINE_TYPE" node="NNN.NNN.NNN.NNN" result=null serial="SERIAL"
2021-11-02T02:19:03.222647Z ubuntu-local neco error: "unknown machine type"
```

morimoto-cybozu (approved):
LGTM

PR #1833, opened 2021-11-16
Update CKE to 1.21.2

morimoto-cybozu:

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -1,7 +1,7 @@
 # tool versions used for both local bin and deb/zip packages
 
 ## These should be updated when Kubernetes is updated
-CKE_VERSION = 1.21.1
+CKE_VERSION = 1.21.2
 CONTAINERD_VERSION = 1.5.6
 CRITOOLS_VERSION = 1.21.0
 RUNC_VERSION = 1.0.2


yokaze:
CKE 1.21.2 is not yet released due to CI error.
https://github.com/cybozu-go/cke/runs/4221748542?check_suite_focus=true

morimoto-cybozu:
@yokaze 
> CKE 1.21.2 is not yet released due to CI error. https://github.com/cybozu-go/cke/runs/4221748542?check_suite_focus=true

Thanks for the notification.  I'll fix it.
Neco CI and release are not affected by the failure of the build of CKE assets because Neco builds CKE binaries from the git repository.
So please make a review of this PR.

yokaze:
@morimoto-cybozu 
I think it's better to release CKE first, because the failing CI may indicate some kind of problem affecting the real cluster.
If there is some kind of proof that the CI failure will not affect the cluster, then it would be OK to go forward.

morimoto-cybozu:
@yokaze 
> If there is some kind of evidence that the CI failure will not affect the cluster, then it would be OK to go forward.

Neco CI confirms that the cluster works well with CKE 1.21.2.

morimoto-cybozu:
I suppose that the cause of the CKE CI failure would be the `flatcar-stable` disk imge used by the CKE sonobuoy test.
The environment of the CKE sonobuoy test is different from that of the Neco CI, as well as from the real environment.

yokaze (approved):
LGTM

Issue #1817, opened 2021-11-11
Retry several times when downloading tools.


kmdkuk:
## What

- When downloading tools during testing, a 503 error returned from GitHub causes the test to fail in an unintended location.

## How

- Add retry handling in Makefile.tools.

## Checklist

- [ ] Finish implementation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #1844, opened 2021-11-22
Add '--retry-on-http-error=503' to wget options

kmdkuk:
- When executing wget, 503 status codes were returned from the download destination, and `make -f Makefile.tools` often failed.
- Fixed to retry wget when a 503 status code comes back.

#1817 

### DIFF
--- a/Makefile.tools
+++ b/Makefile.tools
@@ -2,7 +2,7 @@ include Makefile.common
 
 MAKEFLAGS += -j$(shell nproc)
 CURL=curl -Lsf
-WGET=wget --retry-connrefused --no-verbose
+WGET=wget --retry-on-http-error=503 --retry-connrefused --no-verbose
 
 ## Cache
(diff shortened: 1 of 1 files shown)


yokaze (commented):
Does it wait for several seconds between retries?

kmdkuk:
@yokaze 
Each failure increases the time to wait by one second.
The maximum waiting time can be specified with the `--waitretry=seconds` option, the default is 10 seconds.
Checking `man wget`, I found the following explanation.

```
       --waitretry=seconds
           If you don't want Wget to wait between every retrieval, but only between retries of failed downloads, you can use this option.  Wget will use linear backoff, waiting 1 second after the first failure on a given
           file, then waiting 2 seconds after the second failure on that file, up to the maximum number of seconds you specify.

           By default, Wget will assume a value of 10 seconds.
```

By the way, the default number of retries is 20.

yokaze (approved):
LGTM

PR #1847, opened 2021-11-25
Run chrony-monitor.timer after chrony-wait.service

yokaze:
Previously, `chrony-monitor.service` did not run after `chrony-wait.service`.
This PR fixes the order.


ymmt2005 (approved):
LGTM

PR #1850, opened 2021-11-29
dctest: Run chrony on VM

ysksuzuki:
This PR modifies placemat-menu to create a VM and run a chrony process on it so that the chrony won't disrupt its host's system clock. Also, it updates flatcar.


menu/testdata/setup-iptables
```diff
@@ -3,4 +3,5 @@
 # eth0 -> internet
 # eth1 -> BMC
 iptables -t nat -A POSTROUTING -o eth0 -s 172.19.0.0/24 -j MASQUERADE
+iptables -t nat -A POSTROUTING -o eth0 -s 172.16.4.64/29 -j MASQUERADE
```
zoetrope, line 6: Could you add a comment `# eth0 -> ntp` ?

ysksuzuki: Sure. 

This rule is necessary for chrony VM to access the Internet. eth0 in the core switch connects pseudo internet 10.0.0.0/24, and packets destined to the internet need to be SNAT with eth0 `10.0.0.2/24` here, because those packets are supposed to be SNAT with the global IP on the host side by the following rule.

```
Chain PLACEMAT (1 references)
target     prot opt source               destination
MASQUERADE  all  --  10.0.0.0/24         !10.0.0.0/24
```

zoetrope (approved):

Issue #1867, opened 2021-12-09
neco session-log start crashes for deleted user


umezawatakeshi:
**Describe the bug**
If the login user is deleted when login shell spawned from `neco session-log start` terminates, the `neco` command crashes.

**To Reproduce**
1. login to boot server by some user. user shell runs under `neco session-log start`.
2. delete the user by etcdpasswd.
3. logout

PR #1868, opened 2021-12-09
fix a crash when the current user is unknown.

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1867


masa213f (approved):
LGTM

zoetrope (approved):
LGTM

PR #1907, opened 2022-01-31
Add new NICs for post-boot script

morimoto-cybozu:


masa213f (approved):
LGTM

PR #1915, opened 2022-02-07
Apply the machine-type of the new equipment.

kmdkuk:


pkg/generate-sabakan-machines/main.go
```diff
@@ -26,14 +26,14 @@ func main() {
 	var machineTypeBoot, machineTypeCS, machineTypeSS, bmcType string
 	switch os.Args[1] {
```
zoetrope, line 27: Do we need this switch statement?

How about passing the machine type as a cli option?

kmdkuk: Yes, it is. That way, We don't have to modify the program every time a new equipment appears.

kmdkuk: @zoetrope Modified to accept machine-type as a command line argument using cobra.
The logic for creating json has not been changed.
Please review again.

pkg/generate-sabakan-machines/cmd/root.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:          "generate-sabakan-machines CSV",
+	SilenceUsage: true,
+	Short:        "generate machines.json",
+	Long: `Generate machines.json from the CSV file for use with the 'sabactl create' command.
+Example:
+  generate-sabakan-machines CSV --machine-type-boot=r6525-boot-1 --machine-type-cs=r6525-cs-1 --machine-type-ss=r7525-ss-1
```
zoetrope, line 22: `CSV` does not look like a csv file.

```suggestion
  generate-sabakan-machines input.csv --machine-type-boot=r6525-boot-1 --machine-type-cs=r6525-cs-1 --machine-type-ss=r7525-ss-1
```

zoetrope (approved):

zoetrope (approved):

PR #1916, opened 2022-02-08
Set ManageForeignRoutingPolicyRules and ManageForeignRoutes no

ysksuzuki:
Networkd drops foreign routing policy rules by default from systemd v248.
https://github.com/systemd/systemd/commit/0b81225e5791f660506f7db0ab88078cf296b771

We need to set ManageForeignRoutingPolicyRules=no and ManageForeignRoutes=no to protect foreign routing policy rules, otherwise networkd drops the rules created by Coil and Cilium. It leads to network disruption when one of the underlying interfaces on a node goes down.


morimoto-cybozu (approved):
LGTM

PR #1936, opened 2022-02-25
Distribute Neco operation CLI commands for Teleport

morimoto-cybozu:
This commit enables distribution of Neco operation CLI commands
via Sabakan to worker nodes.
The distributed commands are to be used by Teleport node Pods.


yokaze (approved):
LGTM

PR #1943, opened 2022-03-08
Enable Cilium L4LB

ysksuzuki:
This PR includes
- Enable Cilium kube-proxy replacement
- Update cke-template.yaml to disable kube-proxy
- Separate the cilium manifest and cke-template for stage and production so that we can apply L4LB only on stage


zoetrope (commented):
Isn't it necessary to call `make CILIUM_PRE=true update-cilium` in `check-generate`?

cilium/pre/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+policyAuditMode: false
+policyEnforcementMode: "default"
+prometheus:
+  enabled: true
+rollOutCiliumPods: true
+resources:
+  requests:
+    cpu: 100m
+    memory: 400Mi
+tunnel: "disabled"
```
zoetrope, line 73: I prefer to have `values.yaml` sorted to compare `pre/values.yaml` and `prod/values.yaml`.
You can use `yq e 'sort_keys(.)' values.yaml `

progs/cke/resource.go
```diff
@@ -24,8 +25,33 @@ func UpdateResources(ctx context.Context) error {
 	for _, img := range images {
 		templateParams["cke-"+img.Name] = img.FullName(false)
 	}
+	lbAddr, err := st.GetLBAddressBlockDefault(ctx)
```
zoetrope, line 28: This error has been ignored.

progs/cke/resource.go
```diff
@@ -51,3 +77,15 @@ func UpdateResources(ctx context.Context) error {
 	}
 	return nil
 }
+
+func setLBAddress(key, lbAddr string, templateParams map[string]string, err error) (map[string]string, error) {
+	switch err {
+	case storage.ErrNotFound:
+		templateParams[key] = ""
```
zoetrope, line 84: Is this correct?
What happens if we forget to run `neco config set lb-address-block-*`?

ysksuzuki: We can't throw an error if storage.ErrNotFound occurs because lb-address-block-* is always empty when we upgrade neco that includes this PR. Otherwise the neco upgrade will abort.

Once we upgrade the neco, bgp-cm configmap will be created with empty lb-address-block-*, we will set lb-address-block-*, run `neco init-data`.  Then the bgp-cm configmap will be correct. We can restart cilium-agents then.

cilium/pre/values.yaml
ysksuzuki: `yq e 'sortKeys(.)' values.yaml` ?

zoetrope (dismissed):

zoetrope (approved):

Issue #1684, opened 2021-07-02
CPU pinning


zoetrope:
## Description

Provide CPU pinning mechanism.

## Initiative / Motivation

MySQL may cause serious performance problems.

## Technical information

- https://github.com/intel/CPU-Manager-for-Kubernetes

## Scope and goal

Out of scope:
- Considering NUMA ( https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/ ) 

## Definition of ready

- [ ] The epic owner is assigned.
- [ ] Initiatives and scopes are clearly stated.
- [ ] Feasibility is proved by PoC or scrutiny.
- [ ] All necessary issues that complete this epic are created.
- [ ] At least three people agree on making this ready.

PR #1963, opened 2022-04-05
enable cpu pinning

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1684

- change CPU manager policy to `static`
- reserve some CPUs for system
- remove kubelet state files before it starts


morimoto-cybozu (approved):
LGTM

Issue #1978, opened 2022-04-14
Wait for NICs to have expected IP addresses


morimoto-cybozu:
## What

Neco provides a generator script of systemd-networkd configuration files for worker nodes as shown in the following.
https://github.com/cybozu-go/neco/blob/release-2022.04.12-17983/ignitions/common/files/opt/sbin/setup-local-network

This works fine if the network device names are as expected.
If the network device names differ from expectation, however, it may leave DHCP-assigned addresses.

A worker node should wait for the network devices to have expected IP addresses.

## How

Describe how to address the issue.

## Checklist

- [x] Finish implementation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1982, opened 2022-04-15
wait for network interfaces to have expected addresses

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1978

### DIFF
--- a/ignitions/common/files/opt/sbin/setup-local-network
+++ b/ignitions/common/files/opt/sbin/setup-local-network
@@ -69,3 +69,25 @@ Address={{ index .Spec.IPv4 0 }}/32
 EOF
 
 systemctl restart systemd-networkd.service
+
+sleep 5
+
+wait_network() {
+    DESC=$1
+    EXP=$2
+    shift; shift
+    while true; do
+        for i in $@; do
+            ADDR=$(ip -j addr show $i 2>/dev/null | jq -r '.[].addr_info[]|select(.family=="inet")|.local+"/"+(.prefixlen|tostring)')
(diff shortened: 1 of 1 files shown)


ignitions/common/files/opt/sbin/setup-local-network
```diff
@@ -69,3 +69,26 @@ Address={{ index .Spec.IPv4 0 }}/32
...
+    EXP=$2
+    shift; shift
+    while true; do
+        ADDR=""
+        for i in $@; do
+            ADDR=${ADDR}$(ip -j addr show $i 2>/dev/null | jq -r '.[].addr_info[]|select(.family=="inet")|.local+"/"+(.prefixlen|tostring)')
+        done
+        if [ "${ADDR}" == "${EXP}" ]; then
+            break
+        fi
```
morimoto-cybozu, line 86: Rather than appending to `ADDR`, you can use `break 2` or `return` in a nested loop.
`=` would be the canonical operator when comparing two strings in `[ ... ]`.

```suggestion
        for i in $@; do
            ADDR=$(ip -j addr show $i 2>/dev/null | jq -r '.[].addr_info[]|select(.family=="inet")|.local+"/"+(.prefixlen|tostring)')
            if [ "${ADDR}" = "${EXP}" ]; then
                return
            fi
        done
```

morimoto-cybozu (changes requested):

morimoto-cybozu (dismissed):
LGTM

morimoto-cybozu (approved):
LGTM

PR #1985, opened 2022-04-18
Check auto-generated files

masa213f:
Check auto-generated files in CI. And re-generate manifests.

1. The API version of CronJob in cillium manifests are changed from `batch/v1beta1` to `batch/v1` with the upgrade of the helm command.
The API version is changed by the build-in parameter  (`.Capabilities.KubeVersion.Version`) of the helm command.
https://github.com/cilium/cilium/blob/v1.11.4/install/kubernetes/cilium/templates/hubble/tls-cronjob/cronjob.yaml#L2
https://github.com/cilium/cilium/blob/v1.11.4/install/kubernetes/cilium/templates/_helpers.tpl#L83-L92

2. The coil manifest may be updated manually in the following PR. So I update the source file.
https://github.com/cybozu-go/neco/pull/1980.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ jobs:
     steps:
       - checkout
       - run: time make SUDO="" setup
-      - run: time make check-coil-version
+      - run: time make check-generate
       - run: time make test
       - setup-tools
       - run: time make deb SUDO="" FAKEROOT=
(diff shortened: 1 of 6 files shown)


yokaze (commented):
Is there any reason for not using a strategic merge patch for coil-controller Deployment?

masa213f:
@yokaze 
The strategic merge patch cannot add elements to the tolerations array as far as I tried.

ysksuzuki (approved):
Thank you for creating the patch for coil-controller! I should have done that in the first place.

yokaze (approved):
LGTM

PR #1986, opened 2022-04-20
Use policy/v1 PodDisruptionBudget

masa213f:
The policy/v1beta1 PodDisruptionBudget is deprecated. So I use `policy/v1`.

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125


yokaze (approved):
This PR is LGTM, while `etc/coil.yaml` is still using `policy/v1beta1`. I suspect you already know it.

masa213f:
`etc/coil.yaml` is auto-generated files.
The upstream (https://github.com/cybozu-go/coil) is using policy/v1beta1 PSP, so I'm keeping it.

PR #1992, opened 2022-04-25
Export statuses of several systemd services on boot servers

morimoto-cybozu:

### DIFF
--- a/debian/DEBIAN/postinst
+++ b/debian/DEBIAN/postinst
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 
 SERVICES="node-exporter neco-updater neco-worker sabakan-state-setter cke cke-localproxy teleport-node"
-TIMERS="docker-prune kill-old-login-sessions backup-cke-etcd"
+TIMERS="docker-prune kill-old-login-sessions backup-cke-etcd export-unit-status"
 
 configure() {
     # temporary code for transition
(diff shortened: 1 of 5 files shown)


debian/usr/sbin/export-service-status
```diff
@@ -0,0 +1,18 @@
...
+
+for service in ${SERVICES}; do
+    if systemctl is-active ${service}.service; then
+        value=1
+    else
+        value=0
+    fi
+    name=systemd_$(echo ${service} | tr - _)_is_active
+    echo "# TYPE ${name} gauge" > ${DIR}/${name}.prom.$$
+    echo "${name} ${value}" >> ${DIR}/${name}.prom.$$
```
umezawatakeshi, line 16: how about `systemd_unit_active{unit="neco-updater.service"}` instead of embedding unit name into metric name ?

morimoto-cybozu: @umezawatakeshi 
I also considered that option, and chose not to use labels according to the following criteria.
Now I'm reconsidering, however, because `min(systemd_unit_active)` is not useful but meaningful to some extent.

https://prometheus.io/docs/practices/naming/
> As a rule of thumb, either the `sum()` or the `avg()` over all dimensions of a given metric should be meaningful (though not necessarily useful).

umezawatakeshi (approved):

umezawatakeshi (approved):

PR #2008, opened 2022-05-20
Promtail v2.5.0

masa213f:
Update promtail to v2.5.0.

I change the `clients` config instead of the `client` config, in addition to updating the image.
The `client` config seems to be deprecated.

https://github.com/grafana/loki/blob/v2.5.0/clients/pkg/promtail/config/config.go#L22-L24
https://github.com/grafana/loki/pull/536


yokaze (approved):
LGTM

PR #2011, opened 2022-05-20
Run kubelet image GC earlier than hard eviction

morimoto-cybozu:


umezawatakeshi (approved):

umezawatakeshi:
By the way, How should the order of config parameters (except apiVersion and kind) be?
- just alphabetically sorted
- grouped by function

umezawatakeshi (approved):

PR #2016, opened 2022-06-06
Add nerdctl to worker

zeroalphat:
This is the configuration for deploying nerdctl binaries to neco workers.
Signed-off-by: zeroalphat [yotiyotiarukinopengin@gmail.com](mailto:yotiyotiarukinopengin@gmail.com)

PR #2018, opened 2022-06-13
Add nerdctl local config file

zeroalphat:
part of #2016 
Signed-off-by: zeroalphat <yotiyotiarukinopengin@gmail.com>


zoetrope (commented):
Please describe this PR.
For example, 
- This is a required setting to retrieve container images from nerdctl local registry.
- This setting is for PoC, so it will be removed in the future.

PR #2034, opened 2022-07-15
node-distribute internet-egress Pods

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/2036

The original manifest has soft inter-rack distribution. However, multiple replicas can run on same node and their eviction will fail due to its corresponding PDB. Hard inter-node distribution should be configured.

### DIFF
--- a/etc/squid.yml
+++ b/etc/squid.yml
@@ -191,6 +191,14 @@ spec:
           - 127.0.0.1
       affinity:
         podAntiAffinity:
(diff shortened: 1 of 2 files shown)


ysksuzuki:
As you might have already been aware, please update this as well. 
https://github.com/cybozu-private/neco-apps/blob/main/Makefile#L75

ysksuzuki (approved):

PR #2061, opened 2022-08-22
add option to wait for stop to `neco power` subcommand.

umezawatakeshi:

### DIFF
--- a/pkg/neco/cmd/power.go
+++ b/pkg/neco/cmd/power.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"os/exec"
+	"time"
 
 	"github.com/cybozu-go/log"
 	"github.com/cybozu-go/neco"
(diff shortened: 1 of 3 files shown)


pkg/neco/cmd/power.go
```diff
@@ -97,6 +99,23 @@ func getComputerSystem(service *gofish.Service) (*redfish.ComputerSystem, error)
 	return systems[0], nil
 }
 
+func waitForStop(ctx context.Context, nodeAddr string) error {
+	// It might be better to use any golang library instead of executing ping command...
```
morimoto-cybozu, line 103: Using the ping command would be a good choice because we need an extra privilege for sending raw ICMP packets.
Other choices are to send TCP/UDP packets, but they can be blocked/dropped by various reasons.

pkg/neco/cmd/power.go
```diff
@@ -168,6 +201,9 @@ var powerCmd = &cobra.Command{
 	},
 }
 
+var waitForStopFlag bool
+
 func init() {
 	rootCmd.AddCommand(powerCmd)
+	powerCmd.Flags().BoolVar(&waitForStopFlag, "wait-for-stop", false, "")
```
morimoto-cybozu, line 208: @umezawatakeshi 
If this is not a temporary modification, please add a good description and update `docs/neco.md`.

morimoto-cybozu (approved):
LGTM

PR #2092, opened 2022-09-15
Ignore error in kill-old-login-sessions

masa213f:
Sometimes, kill-old-login-sessions.service is failed with the following error.
It seems that a session may disappear immediately after listing session IDs.
This PR ignores the error.

```
$ journalctl -u kill-old-login-sessions.service
-- Logs begin at Wed 2022-09-14 23:08:01 UTC, end at Thu 2022-09-15 07:20:54 UTC. --
Sep 15 00:00:00 gcp0-boot-2 systemd[1]: Starting kill old login sessions...
Sep 15 00:00:01 gcp0-boot-2 kill-old-login-sessions[57632]: Failed to get session path: No session '42' known
Sep 15 00:00:01 gcp0-boot-2 systemd[1]: kill-old-login-sessions.service: Main process exited, code=exited, status=1/FAILURE
Sep 15 00:00:01 gcp0-boot-2 systemd[1]: kill-old-login-sessions.service: Failed with result 'exit-code'.
Sep 15 00:00:01 gcp0-boot-2 systemd[1]: Failed to start kill old login sessions.
```

### DIFF
--- a/debian/usr/sbin/kill-old-login-sessions
+++ b/debian/usr/sbin/kill-old-login-sessions
@@ -1,18 +1,16 @@
 #!/bin/sh -e
 
-limit=86400 # 1 day
+limit="1day"
 now=$(date +%s)
 
-sessions=$(loginctl list-sessions --no-legend | sed -e 's/^ *//g' | cut -f 1 -d ' ')
-for session in $sessions
-do
-  timestamp=$(loginctl show-session -p Timestamp --value $session)
-  target=$(date +%s --date "$timestamp")
-  diff=$(( $now - $target ))
-  if [ $diff -gt $limit ]; then
+sessions=$(loginctl list-sessions -o json | jq -r ".[].session")
+for session in $sessions; do
+  timestamp=$(loginctl show-session -p Timestamp --value $session) || continue
+  expirationtime=$(date +%s --date "$timestamp + $limit")
+  if [ $now -gt $expirationtime ]; then
     echo "kill session:"
-    loginctl show-session $session
+    loginctl show-session $session || continue
     echo
-    loginctl kill-session --signal 9 $session
+    loginctl kill-session --signal 9 $session || continue
   fi
 done


debian/usr/sbin/kill-old-login-sessions
```diff
@@ -3,10 +3,13 @@
 limit=86400 # 1 day
 now=$(date +%s)
 
-sessions=$(loginctl list-sessions --no-legend | sed -e 's/^ *//g' | cut -f 1 -d ' ')
-for session in $sessions
-do
-  timestamp=$(loginctl show-session -p Timestamp --value $session)
+sessions=$(loginctl list-sessions -o json | jq -r ".[].session")
+for session in $sessions; do
+  timestamp=$(loginctl show-session -p Timestamp --value $session || true)
```
morimoto-cybozu, line 8: The session listed above may disappear everywhere in the loop.
`loginctl show-session` and `loginctl kill-session` could also return errors, so it would be good to ignore errors completely, as you have stated in another place.
An error in this program should not be fatal because this is not a mandatory job.

morimoto-cybozu (changes requested):

debian/usr/sbin/kill-old-login-sessions
morimoto-cybozu: > An error in this program should not be fatal

An error is not fatal but it can cause a failure of `kill-old-login-sessions.service` if happens at the end of the loop, and finally it can cause a misjudgment about the machine's health.
One solution is to add `|| true` after every `loginctl` in the loop.
Another is to configure the `.service` unit to ignore an error, though I don't know how.

masa213f: Thank you.
I fixed to continue the loop with ` || continue` when `loginctl` failed.

If the `date` command fails, this script will fail as follows.

```console
$ ./debian/usr/sbin/kill-old-login-sessions 
date: invalid date ‘Fri 2022-09-16 05:41:36 UTC + 1da’
$ echo $?
1
```

```patch
diff --git a/debian/usr/sbin/kill-old-login-sessions b/debian/usr/sbin/kill-old-login-sessions
index 5cf03238..7d700c47 100755
--- a/debian/usr/sbin/kill-old-login-sessions
+++ b/debian/usr/sbin/kill-old-login-sessions
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 
-limit="1day"
+limit="1da"
 now=$(date +%s)
 
 sessions=$(loginctl list-sessions -o json | jq -r ".[].session")
```

morimoto-cybozu (approved):
LGTM

kmdkuk (approved):
LGTM

PR #2093, opened 2022-09-15
Fix parsing rack number

zoetrope:

### DIFF
--- a/pkg/generate-sabakan-machines/cmd/root.go
+++ b/pkg/generate-sabakan-machines/cmd/root.go
@@ -75,7 +75,7 @@ Example:
 					datacenter, rackString, serial, product, role)
 			}
 
-			rack, err := strconv.ParseInt(rackString, 30, 64)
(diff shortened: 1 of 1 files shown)


pkg/generate-sabakan-machines/cmd/root.go
```diff
@@ -75,7 +75,7 @@ Example:
 					datacenter, rackString, serial, product, role)
 			}
 
-			rack, err := strconv.ParseInt(rackString, 30, 64)
+			rack, err := strconv.ParseInt(rackString, 10, 64)
```
morimoto-cybozu, line 78: I don't think `ParseInt()` is necessary because we don't need int64.

```suggestion
			rack, err := strconv.Atoi(rackString)
```

morimoto-cybozu (changes requested):

morimoto-cybozu (dismissed):
LGTM

morimoto-cybozu (approved):
Rebased to the latest main branch.

PR #2094, opened 2022-09-15
change parameters for unbound tuning

terassyi:
I changed `etc/unbound.yml` and `ignitions/common/files/etc/sysctl.d/70-cybozu.conf`.
This PR changes unbound configurations and sysctl parameters to tune unbound.

### DIFF
--- a/etc/unbound.yml
+++ b/etc/unbound.yml
@@ -169,10 +169,6 @@ spec:
       # other than kube-system as of k8s 1.12.
       # see https://github.com/kubernetes/kubernetes/pull/65593
       #priorityClassName: system-cluster-critical
-      securityContext:
-        sysctls:
-        - name: net.core.rmem_default
-          value: "4194304"
       tolerations:
         - key: node-role.kubernetes.io/master
           effect: NoSchedule
--- a/etc/unbound.yml
+++ b/etc/unbound.yml
@@ -37,11 +37,11 @@ data:
       minimal-responses: yes
       extended-statistics: yes
       rrset-cache-size: 256m
-      rrset-cache-slabs: 2
+      rrset-cache-slabs: 4
       msg-cache-size: 256m
-      msg-cache-slabs: 2
-      infra-cache-slabs: 2
-      key-cache-slabs: 2
+      msg-cache-slabs: 4
+      infra-cache-slabs: 4
+      key-cache-slabs: 4
       so-rcvbuf: 4m
       pidfile: "/tmp/unbound.pid"
       infra-host-ttl: 60
--- a/etc/unbound.yml
+++ b/etc/unbound.yml
@@ -28,11 +28,21 @@ data:
       log-replies: yes
       log-local-actions: yes
       log-servfail: yes
+      num-threads: 4
+      incoming-num-tcp: 2240
+      outgoing-num-tcp: 1000
+      num-queries-per-thread: 4096
+      outgoing-range: 8192
       rrset-roundrobin: yes
       minimal-responses: yes
       extended-statistics: yes
       rrset-cache-size: 256m
+      rrset-cache-slabs: 2
       msg-cache-size: 256m
+      msg-cache-slabs: 2
+      infra-cache-slabs: 2
+      key-cache-slabs: 2
+      so-rcvbuf: 4m
       pidfile: "/tmp/unbound.pid"
       infra-host-ttl: 60
       prefetch: yes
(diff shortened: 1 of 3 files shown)


ignitions/common/files/etc/sysctl.d/70-cybozu.conf
```diff
@@ -12,6 +12,9 @@ net.ipv4.neigh.default.gc_thresh3 = 4096
 net.ipv4.tcp_keepalive_time = 600
 # Increase the maximum TCP send buffer size settable with setsockopt(2).
 net.core.wmem_max = 16777216
+# Increase the maximum TCP receive buffer size
+net.core.rmem_max = 4194304
+rmem_default = 4194304
```
ymmt2005, line 17: Did you try this on your Ubuntu?
It seems that the correct name is `net.core.rmem_default`.
https://docs.kernel.org/admin-guide/sysctl/net.html?highlight=rmem_default#proc-sys-net-core-network-core-options

terassyi: I'm sorry, but I forgot to check it.

etc/unbound.yml
```diff
@@ -28,11 +28,21 @@ data:
...
+      num-threads: 4
+      incoming-num-tcp: 2240
+      outgoing-num-tcp: 1000
+      num-queries-per-thread: 4096
+      outgoing-range: 8192
       rrset-roundrobin: yes
       minimal-responses: yes
       extended-statistics: yes
       rrset-cache-size: 256m
+      rrset-cache-slabs: 2
```
ysksuzuki, line 40: Would you be able to see https://github.com/cybozu-go/cke/pull/567#discussion_r974869584 and modify these parameters accordingly?

ymmt2005 (commented):
@terassyi @ysksuzuki 
I'm not pretty sure about these settings being applied for all services.
Aren't they specific to Unbound?

If so, I'd suggest customizing sysctl values only for Unbound pods.
https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#setting-sysctls-for-a-pod

etc/unbound.yml
terassyi: Should we change `incoming-num-tcp: 1000` according to its pull request?

https://github.com/cybozu-go/cke/blob/2f299cb7bf6bb689b71889a442fcaafa4261883b/op/nodedns/nodedns.go#L36

ysksuzuki:
@ymmt2005 
Do you mean that it might be better to apply the changes of the parameters `net.core.rmem_max` and `net.core.rmem_default` only to unbound pods? If so, I think we probably should do it, because we change these parameters for unbound following the instruction https://unbound.docs.nlnetlabs.nl/en/latest/topics/performance.html

etc/unbound.yml
ysksuzuki: I don't think so. Would you tell me why do think we need to change it?

ymmt2005:
Correct. But chances are there that those settings might benefit other services in general.
Applying the settings only to Unbound pods is safe, anyway.

etc/unbound.yml
terassyi: I also don't think that we need to change it.
I just wanted to confirm if we had to change.

ysksuzuki:
@ymmt2005 We also run unbound containers in node-dns pods and those pods are `hostNetwork: true`. If we apply the changes of net.core.rmem_max and net.core.rmem_default for node-dns as well, I think we need to change them globally. What do you think?

ymmt2005:
@ysksuzuki 
Unbound for node-local DNS serves only for a few pods and receives requests over UDP.
So, these settings won't make sense.

ysksuzuki:
Right, those extend the TCP  receive buffer size. It doesn't make sense then.

ymmt2005:
@ysksuzuki @terassyi 
I revisited these parameters and thought that having `net.core.rmem_max = 4194304` as a global setting is worth doing.
This setting does allow applications to have a 4MiB receive buffer at maximum, which is good for modern network apps.

Setting the default to 4 MiB would waste much memory, I think.

ymmt2005:
@ysksuzuki correct!

ymmt2005 (approved):
LGTM.
I'd suggest telling other teams about the settings for larger receive buffer size might be worth considering if they are running a busy network server such as Nginx.

ysksuzuki (approved):

PR #2100, opened 2022-09-20
serf: Set broadcast timeout to one minute

yokaze:
Serf fails to update tags without a longer broadcast timeout in a large cluster.
https://github.com/hashicorp/consul/issues/8435

This PR extends the timeout to one minute.


masa213f (approved):
LGTM

PR #2108, opened 2022-10-03
Extend CT lifetimes for non-TCP

ysksuzuki:
This PR extends CT lifetimes for non-tcp connections to fix an issue that long-life connections through coil-egress nat are disrupted. 

It seems that Cilium helm chart doesn't support bpf-ct-timeout-*, so I patched the cilium-config cm generated with `helm template` instead. We need to restart cilium-agent manually because the check-sum in the cilium ds manifest is not modified.

### DIFF
--- a/cilium/pre/configmap.yaml
+++ b/cilium/pre/configmap.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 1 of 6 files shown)


yokaze (approved):
LGTM

terassyi (approved):
LGTM

PR #2111, opened 2022-10-06
Use extraConfig

ysksuzuki:
This PR sets bpf-ct-timeout* using extraConfig instead of the cm patch.
https://github.com/cilium/cilium/tree/master/install/kubernetes/cilium

### DIFF
--- a/cilium/pre/configmap.yaml
+++ b/cilium/pre/configmap.yaml
@@ -1,8 +1,0 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: cilium-config
-  namespace: kube-system
-data:
-  bpf-ct-timeout-regular-any: 1h0m0s
-  bpf-ct-timeout-service-any: 1h0m0s
--- a/cilium/pre/kustomization.yaml
+++ b/cilium/pre/kustomization.yaml
@@ -4,5 +4,3 @@ resources:
   - ../base
   - bgp_config.yaml
   - upstream.yaml
-patchesStrategicMerge:
-  - configmap.yaml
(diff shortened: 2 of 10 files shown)


terassyi (commented):
Now we can use the `--set-json` option with helm v3.10.0 to configure objects like the below.

```
--set-json='extraConfig={"bpf-ct-timeout-regular-any": "1h0m0s", "bpf-ct-timeout-service-any": "1h0m0s"}' \
```

Please consider using this.

This feature is added by helm v3.10.0 release.
- https://github.com/helm/helm/releases/tag/v3.10.0
- https://github.com/helm/helm/pull/10693

ysksuzuki:
@terassyi Thank you for your comment.
I don't see why we should use `--set-json`. What is the benefit of using `--set-json`?

terassyi (commented):
There are `values.yaml ` for each environment; pre and prod.
And we seem to configure common values with helm --set option.

So I think it may be better that we set environment-specific configurations in values.yaml and use --set-json for common configurations.

ysksuzuki:
@terassyi 
> There are values.yaml for each environment; pre and prod.
> And we seem to configure common values with helm --set option.

We only configure image settings with `--set` option in the Makefile because tags are retrieved from `artifacts.go` in it, e.g. `CILIUM_TAG`. Other than those, we configure parameters with values.yaml.

> So I think it may be better that we set environment-specific configurations in values.yaml and use --set-json for common configurations.

The reason we manage two values.yaml for pre and prod is that we sometimes need to do load tests for a while in our staging environment when we enable features like XDP.  In that case, we only enable it with the pre values.yaml not to stop the production release.  bpf-ct-timeout* options are same in both pre and prod now, but it doesn't mean those are common value. Because if we try a larger CT lifetime in our staging then we modified pre value.yaml.

terassyi (approved):
Thank you for your explanation.

I understood the reason why we use both the --set option and values.yaml and there are two values.yaml in pre and prod directories.

LGTM

yokaze (approved):

PR #2117, opened 2022-10-14
Advertise routes for Loadbalancer service from cs nodes

ysksuzuki:


yokaze (approved):
LGTM

terassyi (approved):

PR #2122, opened 2022-10-20
dctest: dcssh: pass identify file path via command line option

daichimukai:
We cannot use `dcssh` and `dcscp` unless the current working directory is `/path/to/neco/dctest` since the ssh config that these tools use requires the identity file `dctest_key` exists in the current working directory. This commit changes `dcssh` and `dcscp` to pass the absolute path of `dctest_key` to an inner ssh command so that these scripts can be run from everywhere.

### DIFF
--- a/dctest/dcscp
+++ b/dctest/dcscp
@@ -1,6 +1,9 @@
 #!/bin/bash
 
 SUDO=$(which sudo)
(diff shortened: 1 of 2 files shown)


zoetrope (approved):
LGTM

daichimukai:
I rebased to the current main branch.

zoetrope (approved):

PR #2156, opened 2022-11-22
Fix tools extraction not to contaminate shared directory

morimoto-cybozu:


terassyi (approved):
LGTM

PR #2185, opened 2022-12-23
Disable the cgroup v2 auto-mount

ysksuzuki:
The cgroup v2 auto-mount is no longer needed because the cgroup v2 has been enabled in our worker nodes.

https://docs.cilium.io/en/stable/gettingstarted/kubeproxy-free/#kubeproxy-free

> Cilium will automatically mount cgroup v2 filesystem required to attach BPF cgroup programs by default at the path /run/cilium/cgroupv2. To do that, it needs to mount the host /proc inside an init container launched by the DaemonSet temporarily. If you need to disable the auto-mount, specify --set cgroup.autoMount.enabled=false, and set the host mount point where cgroup v2 filesystem is already mounted by using --set cgroup.hostRoot. For example, if not already mounted, you can mount cgroup v2 filesystem by running the below command on the host, and specify --set cgroup.hostRoot=/sys/fs/cgroup.


yokaze (approved):
LGTM

PR #2218, opened 2023-02-14
Update dependencies

morimoto-cybozu:


bin/run-dctest-suite.sh
```diff
@@ -23,6 +23,8 @@ fi
 
 # Run dctest
 cd \${GOPATH}/src/github.com/cybozu-go/neco/dctest
```
zeroalphat, line 25: I think the make setup command in `run-dctest.sh` will set up ginkgo v2.
Can you tell me why I need to re-run the setup from `run-dctest-suite.sh`?

morimoto-cybozu: @zeroalphat 
Because I found that the [upgrade test failed once](https://app.circleci.com/pipelines/github/cybozu-go/neco/7120/workflows/9fc5794c-4e74-4ddb-8030-de6a1319805e/jobs/21664) due to the old ginkgo installation:-)

In the upgrade test, `run-dctest.sh` is executed with `datacenter: staging`.
https://github.com/cybozu-go/neco/blob/release-2023.02.13-21646/.circleci/config.yml#L238
https://github.com/cybozu-go/neco/blob/release-2023.02.13-21646/.circleci/config.yml#L52
https://github.com/cybozu-go/neco/blob/release-2023.02.13-21646/bin/run-dctest.sh#L89
It runs `make setup` in the same git tag as the staging datacenter uses, and installs ginkgo V1.
This behavior is the intended one.
We need to reproduce an old environment as the first stage of the upgrade test.
The ginkgo installation is not updated, however, even if the upgraded environment needs V2.

The re-run of `make setup` is necessary only when we upgrade ginkgo.
I'll revert this re-run after we release the git tag using ginkgo V2.

zeroalphat: Thank you for your detailed explanation.
It has helped me to understand.

zeroalphat (approved):
LGTM

PR #2223, opened 2023-02-22
Fix slow deb package build issue

ysksuzuki:
The default compression type of dpkg-deb 1.21.1 is zstd,
It doesn't work concurrently and leads to the timout error
This PR workaround this issue by spcifiying -zxz explicitly


morimoto-cybozu (approved):
LGTM

PR #2224, opened 2023-02-24
Increase controller-manager's QPS to use while talking with apiserver

zoetrope:
This parameter probably solve performance issues.
ref. https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/


morimoto-cybozu (approved):
LGTM

PR #2234, opened 2023-03-08
Increase fs.aio-max-nr to solve the problem of pods not being able to start

zoetrope:
Oracle and Veritas recommend value to 1048576.

https://sort.veritas.com/public/documents/HSO/2.0/linux/productguides/html/hfo_admin_ubuntu/ch04s03.htm


morimoto-cybozu (approved):
LGTM

PR #2239, opened 2023-03-16
Use apt-get for building installer to validate packages

morimoto-cybozu:


zoetrope (approved):
LGTM

PR #2247, opened 2023-04-03
check the length of none-ss nodes before selecting for reboot

peng225:
When we try to exec `neco reboot-worker` with only ss nodes, the neco command crashes.
This is due to the following check.
https://github.com/cybozu-go/neco/blob/4b3f01eff5d608075493b34277cce32b68119a56/pkg/neco/cmd/reboot_worker.go#L193

Because both `nonssIndex` and `ssIndex` is initialized with 0, the condition holds for the first loop, and the empty slice `nonss` is accessed.

To fix this bug, I added a check for the length of the none-ss nodes.


umezawatakeshi:
memo:

- `neco reboot-worker --role ss` crashes as stated in the description.
- `neco reboot-worker --role cs` works correclty as expected.

umezawatakeshi (approved):
seems fixed

PR #2256, opened 2023-04-10
Update teleport 11.3.10

terassyi:


yokaze (approved):
LGTM, but it should be applied to clusters after the deployment of the Teleport's auth server.

terassyi:
Thank you for notifying.
I edited the title not to merge.

PR #2264, opened 2023-04-14
Update dependencies

ysksuzuki:
Exclude cilium and squid image because they didn't run for some reason. Will fix it later.


yokaze (approved):
Is it OK to use a newer version of Cilium CLI? If it is OK, this is LGTM.

PR #2266, opened 2023-04-17
Upgrade Cilium to 1.12.8

ysksuzuki:
We can't upgrade cilium and squid at the same time. Let's upgrade squid separately

- Fail to start squid pod because cilium hasn't started yet
- Fail to pull new cilium image because squid hasn't started yet
  - Communication through the node port svc(squid svc) fail because cilium can't update the bpf map


morimoto-cybozu (approved):
LGTM

PR #2279, opened 2023-05-02
Refine BIRD configuration parameters

morimoto-cybozu:
This commit updates the BIRD configuration to extend the BFD TX/RX intervals
from 300 ms to 400 ms.
The multiplier is set to 5 in default.
This combination means that a link is not declared as down if the link
failure is shorter than 2 sec.
In other words, the detection of the link failure is delayed by 2 sec.

This commit also shorten the minimum delay between a protocol failure
and automatic restart for the BGP protocols from 60 sec (default) to
3 sec.
This will speed up the recovery from a failure.

See https://bird.network.cz/?get_doc&v=20&f=bird-6.html#ss6.2 for the BFD parameters.
See https://bird.network.cz/?get_doc&v=20&f=bird-6.html#ss6.3 for the BGP parameters.

### DIFF
--- a/ignitions/common/files/etc/bird/bird.conf
+++ b/ignitions/common/files/etc/bird/bird.conf
@@ -8,8 +8,8 @@ protocol direct singles {
 }
 protocol bfd {
     interface "*" {
-       min rx interval 300 ms;
-       min tx interval 300 ms;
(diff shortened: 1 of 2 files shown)


zeroalphat (approved):
LGTM

PR #2308, opened 2023-06-01
update libssl3 package url for installer

umezawatakeshi:
workaround for the environment which is detected "not 22.04"


morimoto-cybozu (approved):
LGTM

PR #2364, opened 2023-09-14
Disable iscsiuio.service

yokaze:
This PR:
- masks `iscsiuio.service` because it is not used
- resets failure status of `iscsiuio.service` when masked, because it causes CI instability.


morimoto-cybozu (approved):
LGTM

PR #2375, opened 2023-09-26
Update cilium to v1.13.7 with patches

terassyi:


ysksuzuki (changes requested):
Please check if we need to update our `values.yaml`, following the upgrade guide. 

I think we need to specify `upgradeCompatibility: "1.12"` here
https://github.com/cybozu-go/neco/blob/main/cilium/pre/values.yaml#L80

ysksuzuki (approved):

PR #2377, opened 2023-09-27
update dependencies

YZ775:
Reguler update

latest version of k8s.io/api is incompatible with k8s.io/kube-openapi@v0.0.0-20221012153701-172d655c2280,
so I added replace statement of that.

Signed-off-by: YZ775 <yuzuki-mimura@cybozu.co.jp>


zeroalphat (approved):
LGTM

PR #2379, opened 2023-09-28
reset failed status of systemd-networkd-wait-online.service

umezawatakeshi:
Since yesterday, systemd-networkd-wait-online.service fails due to timeout during neco-apps bootstrap. (The reason is still unknown)
The failure itself doesn't seem a problem because the test code can already connect to boot servers.
Because neco-apps has a alert rule that checks all systemd services are OK, I decide to reset the failed status instead of ignoring the failed status during the test.

note: it may or may not be a temporary resolution.

### DIFF
--- a/dctest/init_test.go
+++ b/dctest/init_test.go
@@ -12,6 +12,14 @@ import (
 
 // testInit test initialization steps
 func testInit() {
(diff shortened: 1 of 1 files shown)


yokaze (approved):
LGTM

PR #2394, opened 2023-10-13
Sabakan TLS step1

YZ775:


constants.go
```diff
@@ -139,6 +139,9 @@ var (
 	ServerCertFile = filepath.Join(NecoDir, "server.crt")
 	ServerKeyFile  = filepath.Join(NecoDir, "server.key")
 
+	SabakanTLSCertFile = filepath.Join(SabakanDir, "sabakan-tls.crt")
+	SabakanTLSKeyFile  = filepath.Join(SabakanDir, "sabakan-tls.key")
```
masa213f, line 143: Please make it clear that they are server certificates.

```suggestion
	SabakanServerCertFile = filepath.Join(SabakanDir, "sabakan-server.crt")
	SabakanServerKeyFile  = filepath.Join(SabakanDir, "sabakan-server.key")
```

pkg/neco/cmd/init_local.go
```diff
@@ -72,7 +84,15 @@ new a application NAME.`,
...
-				err = neco.StartService(ctx, neco.SabakanService)
+				var isActive bool
+				isActive, err = neco.IsActiveService(ctx, neco.SabakanService)
+				if err != nil {
+					return err
+				}
+				if isActive {
+					return neco.RestartService(ctx, neco.SabakanService)
+				}
+				return neco.StartService(ctx, neco.SabakanService)
```
masa213f, line 95: Do we need to restart the sabakan service at this time?
When we will run `neco-init sabakan`, the new sabakan will not be released yet, won't it?

pkg/neco/cmd/init_local.go
```diff
@@ -95,6 +115,10 @@ func init() {
 }
 
 func issueCerts(ctx context.Context, vc *api.Client, commonName, cert, key string) error {
+	_, err := os.Stat(cert)
+	if err == nil {
+		return os.ErrExist
+	}
```
masa213f, line 121: Why did you add this check?

pkg/neco/cmd/init_local.go
```diff
@@ -109,3 +133,32 @@ func issueCerts(ctx context.Context, vc *api.Client, commonName, cert, key strin
 	return neco.WriteFile(key, secret.Data["private_key"].(string))
 
 }
+
+func issueSabakanTLSCerts(ctx context.Context, vc *api.Client, cert, key string) error {
```
masa213f, line 137: Please make it clear that this is for server certificates.

```suggestion
func issueSabakanServerCerts(ctx context.Context, vc *api.Client, cert, key string) error {
```

pkg/neco/cmd/init_data.go
```diff
@@ -74,6 +75,15 @@ func initData(ctx context.Context, st storage.Storage) error {
 		return cke.UpdateResources(ctx, st)
 	}
 
+	if uploadCACertOnly {
+		sabakanClient, err := client.NewClient(neco.SabakanLocalEndpoint, httpClient.Client)
+		if err != nil {
+			return err
+		}
+		_, err = sabakanClient.AssetsUpload(ctx, "sabakan-tls-ca.crt", neco.ServerCAFile, nil)
```
masa213f, line 83: Please create a function in `progs/sabakan` for this logic.

masa213f (changes requested):

pkg/neco/cmd/init_local.go
YZ775: I wrote this code before I came up with the idea of splitting deploy operations to 2 step, so I expected that update of  sabakan and neco command would be same time.
I'll remove this restart logic.

pkg/neco/cmd/init_local.go
YZ775: I want to make `neco-init sabakan` command to be idempotent. so I need to send the existence of the cert file.
This check returns `ErrExist` when the cert file are already exists in specified directry.

masa213f: Are there any side effects of issuing a new cert?

I assume that the `neco init-local` is executed only once during initializing the boot server. So, it is not necessary to be idempotent.

If the side effects of issuing a new cert are negligible, we can overwrite it.

pkg/neco/cmd/init_local.go
```diff
@@ -72,7 +84,14 @@ new a application NAME.`,
 			case "etcdpasswd":
 				err = neco.StartService(ctx, neco.EtcdpasswdService)
 			case "sabakan":
-				err = neco.StartService(ctx, neco.SabakanService)
+				var isActive bool
```
masa213f, line 87: Do you avoid the error by turning off the sabakan service before the operation?
I don't want to leave code that executes only once for updating the sabakan.

masa213f (changes requested):

pkg/neco/cmd/init_local.go
```diff
@@ -52,15 +54,37 @@ new a application NAME.`,
 			var err error
 			switch initLocalParams.name {
 			case "etcdpasswd":
-				err = issueCerts(ctx, vc, "etcdpasswd", neco.EtcdpasswdCertFile, neco.EtcdpasswdKeyFile)
+				err = issueEtcdCerts(ctx, vc, "etcdpasswd", neco.EtcdpasswdCertFile, neco.EtcdpasswdKeyFile)
+				if err != nil {
+					if err != os.ErrExist {
+						return err
+					}
+					fmt.Printf("etcdpasswd certificate already exists in %s.  skipping issue cert\n", neco.SabakanEtcdCertFile)
```
masa213f, line 62:
```suggestion
					fmt.Printf("etcdpasswd certificate already exists in %s.  skipping issue cert\n", neco.EtcdpasswdCertFile)
```

pkg/neco/cmd/init_local.go
```diff
@@ -52,15 +54,37 @@ new a application NAME.`,
...
 				}
 				err = sabakan.InitLocal(ctx, vc)
 			case "cke":
-				err = issueCerts(ctx, vc, "cke", neco.CKECertFile, neco.CKEKeyFile)
+				err = issueEtcdCerts(ctx, vc, "cke", neco.CKECertFile, neco.CKEKeyFile)
+				if err != nil {
+					if err != os.ErrExist {
+						return err
+					}
+					fmt.Printf("cke certificate already exists in %s.  skipping issue cert\n", neco.SabakanEtcdCertFile)
```
masa213f, line 86:
```suggestion
					fmt.Printf("cke certificate already exists in %s.  skipping issue cert\n", neco.CKECertFile)
```

pkg/neco/cmd/init_local.go
```diff
@@ -52,15 +54,37 @@ new a application NAME.`,
...
+				}
 			case "sabakan":
-				err = issueCerts(ctx, vc, "sabakan", neco.SabakanCertFile, neco.SabakanKeyFile)
+				err = issueEtcdCerts(ctx, vc, "sabakan", neco.SabakanEtcdCertFile, neco.SabakanEtcdKeyFile)
 				if err != nil {
-					return err
+					if err != os.ErrExist {
+						return err
+					}
+					fmt.Printf("sabakan certificate already exists in %s.  skipping issue cert\n", neco.SabakanEtcdCertFile)
```
masa213f, line 70: Why don't you use `log.Info()`?

masa213f (changes requested):

masa213f (approved):
LGTM

masa213f (approved):

PR #2400, opened 2023-10-26
Update coil to v2.5.0

terassyi:


terassyi:
marged by https://github.com/cybozu-go/neco/pull/2401

PR #2416, opened 2023-11-15
Update CKE to 1.26.3

zeroalphat:
I was adding the flmegraph and necoperf variables in the wrong place.
I will correct this while upgrading the CKE version.


terassyi (approved):
LGTM

PR #2418, opened 2023-11-16
Downgrade OS image to 3602.2.0

masa213f:
- Downgrade the OS image to `3602.2.0` to investigate network failure.
- Ignore the Squid `6.5.0.1` image to prevent failure when prod release.
  - Cilium `1.13.7.2` will be included in the next prod release.


zeroalphat (approved):
LGTM

PR #2425, opened 2023-11-24
use blob cache to download large files

umezawatakeshi:
part of https://github.com/cybozu-private/neco-tasks/issues/124


bin/download-with-blob-cache.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# If NECO_CI_BLOB_CACHE_URL environment variable is set, this script uses blob cache.
+# Otherwise, this script just download a file.
+
+# Note: cache entry key is generated by <download-url>.
+
+CURLCMD="$1"
+LOCALNAME="$2"
+DLURL="$3"
+CACHEURL=${NECO_CI_BLOB_CACHE_URL%/}/single/$(echo ${DLURL} | sed -E -e 's|^https?://||' -e 's|///*|/|')
```
kmdkuk, line 15: `-e 's|///*|/|'`
What does this mean?
I thought that when `//` appeared in the middle of the path, it would be changed to `/`, but in that case, I thought `'s|///*|/|g'` would be better.

kmdkuk (approved):
LGTM

PR #2435, opened 2023-12-04
increase cs node resources in menu-ss.yml

umezawatakeshi:


zeroalphat (approved):
LGTM

PR #2436, opened 2023-12-04
add debug log configuration for GET pods log

umezawatakeshi:
cf. https://kubernetes.io/docs/reference/config-api/apiserver-audit.v1/


morimoto-cybozu (dismissed):
LGTM

PR #2438, opened 2023-12-05
Set memory limit for serf to mitigate memory leak

yokaze:


zeroalphat (approved):
LGTM

PR #2450, opened 2023-12-14
Expand cs memory for dctest

kmdkuk:
The issue of not being able to increase cpu resources is under investigation.
Memory only seems to be OK, so I increased it.


kmdkuk:
run [dctest-with-neco-feature-branch](https://github.com/cybozu-private/neco-apps/actions/runs/7202942811)

umezawatakeshi (approved):

PR #2452, opened 2023-12-14
container image asset names should include container registry name

umezawatakeshi:
The container image files include their image names. i.e. they include registry name (like "ghcr.io") and username (like "cybozu").
On the other hand, the container image asset names does not include them.

As the result, Neco boot servers and worker nodes may get unexpected image files: an worker node want to get a image from a certain container registry (e.g. `ghcr.io/cybozu/xxx:n.n.n`) but it actually get a image from another container registry (e.g. `quay.io/cybozu/xxx:n.n.n`).

### DIFF
--- a/progs/sabakan/upload.go
+++ b/progs/sabakan/upload.go
@@ -14,6 +14,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"strings"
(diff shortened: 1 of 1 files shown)


morimoto-cybozu (approved):
LGTM

PR #2465, opened 2024-01-15
add squid-exporter

YZ775:


zeroalphat (changes requested):
I thought it would be a good idea to add a test(test like `checkUnboundExporter`) to see if the squid-exporter is working.
https://github.com/cybozu-go/neco/blob/main/dctest/squid_test.go

zeroalphat (approved):
LGTM

PR #2481, opened 2024-01-31
Fix ubuntu cloud image version

zeroalphat:
Since CI uses the latest version of ubuntu-cloud image, there is a possibility that the version of ubuntu-cloud image will be upgraded and CI will break by itself.
To prevent CI from breaking on its own, fix the version of the ubuntu cloud image to be used.


YZ775 (approved):
LGTM

PR #2482, opened 2024-02-01
Update CKE to 1.27.4 to introduce repair queue

morimoto-cybozu:
This PR updates CKE to 1.27.4.
This PR also updates the cluster yaml for the repair queue,
which is introduced in CKE 1.27.4.

### DIFF
--- a/etc/cke-template.yml
+++ b/etc/cke-template.yml
@@ -31,6 +31,49 @@ reboot:
   command_interval: 60
   evict_retries: 30
   evict_interval: 10
+repair:
+  repair_procedures:
+  - machine_types: ["iDRAC"]
+    repair_operations:
+    - operation: unhealthy
+      repair_steps:
+      - repair_command: ["sh", "-c", "timeout 30 neco bmc repair dell reset-idrac $1 || ckecli ssh $1 -- docker exec setup-hw racadm racreset soft", "reset-idrac"]
+        command_timeout_seconds: 60
+        command_retries: 2
+        command_interval: 30
+        watch_second: 600
+      - repair_command: ["neco", "bmc", "repair", "dell", "discharge"]
+        command_timeout_seconds: 30
+        command_retries: 2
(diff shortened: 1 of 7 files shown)


etc/cke-template.yml
```diff
@@ -31,6 +31,49 @@ reboot:
   command_interval: 60
   evict_retries: 30
   evict_interval: 10
+repair:
+  repair_procedures:
+  - machine_types: ["iDRAC"]
+    repair_operations:
+    - operation: unhealthy
+      repair_steps:
+      - repair_command: ["sh", "-c", "timeout 30 neco bmc repair dell reset-idrac $1 || ckecli ssh $1 -- docker exec setup-hw racadm racreset soft", "reset-idrac"]
```
YZ775, line 40: I want to know the purpose of `"reset-idrac"` in the end of the `repair_command[]`, would you explain it?

morimoto-cybozu: @YZ775 
It is a command name for the command string `timeout 30 neco ... || ckecli ...`.

The syntax of `sh -c` is as follows:

```
dash -c command_string [command_name [argument ...]]
```

`reset-idrac` is interpreted as a command name and used when substituting `$0`.
In our use case, we don't use `$0` in the command string, so we could omit the command name and use `$0` to express the practical first argument, the IP address.
Nevertheless I prefer to put a command name explicitly even though unnecessary because it seems unnatural for me to use `$0` as a positional parameter.

YZ775: I didn't know that the first argument of `sh -c` is command_name and it is stored in`$0`.
Thank you for explaining in detail.

I also agree with using command_name and `$1` rather than put IP address on command_name and using `$0`

masa213f (approved):
LGTM

YZ775 (approved):
LGTM

zeroalphat (approved):
LGTM

PR #2489, opened 2024-02-09
Add necoip command

yokaze:
This PR adds `necoip` command, that helps us to find where the IP address comes from.
It will be installed on boot servers and Teleport nodes.


zeroalphat (approved):
LGTM

PR #2503, opened 2024-02-22
Fix clusterdump and verify it

yokaze:
`clusterdump` did not work properly, so I fixed it and added tests.


chez-shanpu (approved):
LGTM

PR #2505, opened 2024-02-28
Disable the image pull progress timeout

zeroalphat:
Change the containerd configuration so that a timeout does not occur when pulling an image.
Set `image_pull_progress_timeout` to `0s` to prevent timeouts.
>A zero value means there is no timeout.
https://github.com/containerd/containerd/blob/v1.7.13/pkg/cri/config/config.go#L374


terassyi (approved):
LGTM

PR #2506, opened 2024-02-29
Add InstallDebianPackage and installDebianPackage's parameter to give Env values for systemd-run

terassyi:
This PR added the parameter(named `envValues`) for `InstallDebianPackage()` and `installDebianPackage()` to set environment values.

Given values in `envValues` will be extracted like following and will be executed.
When `envValues` is given as `map[string]string{"HOME": "/home/cybozu"}`,  these values are extracted like following.

```console
systemd-run -p Environment=HOME=/home/cybozu ...
```

ref: https://github.com/cybozu-go/neco/pull/2506/files#diff-38861c4365e417f67e3b1020fb2297e10cd135406e2913894c4ed32c7f5d2aa4R116

### DIFF
--- a/worker/deb.go
+++ b/worker/deb.go
@@ -18,7 +18,8 @@ import (
 // InstallDebianPackage installs a debian package
 // client uses for downloading a debian package.
 // ghClient uses for getting download URL by GitHub API.
-func InstallDebianPackage(ctx context.Context, client *http.Client, ghClient *github.Client, pkg *neco.DebianPackage, background bool) error {
+// envvalues uses for giving environment values to systemd-run command when background is true.
+func InstallDebianPackage(ctx context.Context, client *http.Client, ghClient *github.Client, pkg *neco.DebianPackage, background bool, envValues map[string]string) error {
 	downloadURL, err := GetGitHubDownloadURL(ctx, ghClient, pkg)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


worker/deb.go
```diff
@@ -66,14 +67,30 @@ func InstallDebianPackage(ctx context.Context, client *http.Client, ghClient *gi
...
 	if background {
 		command = append([]string{"systemd-run", "-q", "--wait"}, command...)
+		if envValues != nil {
+			envs := make([]string, 0, len(envValues)*2)
+			for k, v := range envValues {
+				envs = append(envs, []string{"-p", fmt.Sprintf("Environment=%s=%s", k, v)}...)
+			}
+			command = append(command, envs...)
+		}
 	}
```
YZ775, line 77: I think it's better to place argument of `systemd-run`  before command of `systemd-run`

```suggestion
	command := []string{"sh", "-c", "dpkg -i " + f.Name() + " && rm " + f.Name()}
	if background {
		systemdCommand := []string{"systemd-run", "-q", "--wait"}
		if envValues != nil {
			for k, v := range envValues {
				systemdCommand = append(systemdCommand []string{"-p", fmt.Sprintf("Environment=%s=%s", k, v)}...)
			}
			command := append(systemdCommand, command...)
		}
	}

```

YZ775 (approved):
LGTM

PR #2519, opened 2024-03-15
Add the feature to handle ghcr credentials instead of quay

zeroalphat:


docs/neco.md
```diff
@@ -99,7 +99,7 @@ Synopsis
 
 * `neco join LRN [LRN ...]`
 
-    Prepare certificates and files to add this server to the cluster.
```
yokaze, line 102: The trailing double spaces mean a line break in markdown, so we need to keep it.

yokaze (changes requested):
Could you please also update `docs/etcd.md`?

docs/neco.md
zeroalphat: Formatter had deleted it on its own, so I put it back.
Sorry.

yokaze (approved):
LGTM!

PR #2525, opened 2024-03-21
Fix repair queue config to use appropriate commands

morimoto-cybozu:
* In the case of `unhealthy`, the iDRAC of the unhealthy machine may not
    be working.  So we should skip `neco bmc repair dell reset-idrac`.
* In the case of `unreachable`, SSH to the machine is likely to fail.
    So we should skip `ckecli ssh $1 ...`.
* A repair command in the form of `sh -c '...'` is hard to cancel.
    We should use `sh -c 'exec ...'` if possible.

### DIFF
--- a/etc/cke-template.yml
+++ b/etc/cke-template.yml
@@ -37,13 +37,13 @@ repair:
     repair_operations:
     - operation: unhealthy
       repair_steps:
-      - repair_command: ["sh", "-c", "timeout 30 neco bmc repair dell reset-idrac $1 || ckecli ssh $1 -- docker exec setup-hw racadm racreset soft", "reset-idrac"]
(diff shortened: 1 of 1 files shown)


zeroalphat (dismissed):
LGTM

YZ775 (dismissed):
LGTM

masa213f (approved):

morimoto-cybozu:
@zeroalphat @YZ775 @masa213f 
Thank you for reviewing my PR.
I've updated one of repair commands to use `sh -c 'exec ...'` instead of `sh -c '...'`.
Would you please re-review it?

YZ775 (approved):
LGTM

PR #2535, opened 2024-04-03
Use docker credentials when pulling container images

zeroalphat:
In the current implementation, when pulling container images from neco, the credentials managed by either docker or neco are used.

When pulling container images, docker credentials will be used.
https://github.com/cybozu-go/neco/blob/b13aacd3b261ac6ba648a3d2e19a46839d928f6d/docker.go#L44-L47

When pulling a container image to tarball it, the credentials stored in neco will be used.
https://github.com/cybozu-go/neco/blob/b13aacd3b261ac6ba648a3d2e19a46839d928f6d/image.go#L42

I would like to be able to manage credentials in one of the two, because if I manage credentials in both, there is a possibility that credential update omissions may occur.
This PR will change to use docker credentials when pulling container images to tarball them.

I plan to remove the functions that manage credentials for pulling container images when they are no longer used.

### DIFF
--- a/image.go
+++ b/image.go
@@ -35,11 +35,7 @@ func (f ImageFetcher) GetTarball(ctx context.Context, img ContainerImage, w io.W
 		return err
 	}
 
-	auth := f.auth
-	if auth == nil || !img.Private {
-		auth = authn.Anonymous
-	}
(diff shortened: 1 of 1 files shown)


masa213f (approved):
LGTM

PR #2554, opened 2024-05-01
Make different artifacts.go included in the release

zeroalphat:
Make different artifacts.go included in the release for different environments.


masa213f (commented):
As I commented elsewhere, please consider whether we can stop using the build tag.

.circleci/config.yml
```diff
@@ -262,35 +259,27 @@ jobs:
...
-      - persist_to_workspace:
-          root: .
-          paths:
-            - "artifacts_release.go"
+      - run: make update-artifacts
+      - run: 
+          name: Compare artifacts
+          command: |
+            diffs=$(git diff --name-only main)
+            if [ "$diffs" = "" ]; then printf "%s\n" "$diffs"; exit 0; fi
```
morimoto-cybozu, line 272: The comparison seems inverted.

```suggestion
            diffs=$(git diff --name-only main)
            if [ "$diffs" != "" ]; then printf "%s\n" "$diffs"; exit 0; fi
```

And it could be simplified.

```suggestion
            git --no-pager diff --name-only main
```

I suppose that this `run` step is just for information.
We don't want to stop the sequence even if there were differences, right?

`git diff` without `--exit-code` option returns zero even if there were differences.
You can directly use this command.
In case you didn't know, if `git diff` returned non-zero, the assignment `diffs=$(git diff ...)` also returns non-zero and fails.

`--no-pager` may not be necessary.

.circleci/config.yml
```diff
@@ -429,40 +407,72 @@ workflows:
...
+          name: "build release with release prefix"
+          tags: "release"
           filters:
             branches:
               ignore: /.*/
             tags:
-              only: /^(release|test)-.*/
+              only: /^(release)-.*/
       - deploy_github:
+          name: "build release with release prefix"
```
morimoto-cybozu, line 435:
```suggestion
          name: "deploy release with release prefix"
```

docs/generate-artifacts.md
```diff
@@ -3,7 +3,7 @@ generate-artifacts
 
 `generate-artifacts` is a command-line tool to generate `artifacts.go`
 which is a collection of latest components. This is used while
```
morimoto-cybozu, line 5:
```suggestion
`generate-artifacts` is a command-line tool to generate `artifacts.go`,
which is a collection of the latest components. This tool is used in development.
It is also used in CI to check differences between the latest release and the versions in use.
```

updater/release_checker.go
```diff
@@ -52,7 +52,8 @@ func (c *ReleaseChecker) Run(ctx context.Context) error {
 		github.SetTagPrefix("test-")
 		c.check = github.GetLatestPublishedTag
 	case neco.StagingEnv:
-		c.check = github.GetLatestPublishedTag
+		github.SetTagPrefix("test-")
+		c.check = github.GetLatestReleaseTag
```
morimoto-cybozu, line 56: Do we need to change this?

morimoto-cybozu: At least the following should be updated:

* [`docs/cicd.md`](https://github.com/cybozu-go/neco/blob/release-2024.04.30-28619/docs/cicd.md#ci-flow)
    * The steps of "deploy for the staging data center" should be updated to include the following.
        * Remove the "pre-release" label from the target `test-` release.
    * It is still necessary to create a `release-` tag.  This may be done automatically for a released `test-` release.
    * It currently lacks a description for the dev data center, i.e., stage1.
* [`download-neco-deb`](https://github.com/cybozu-go/neco/blob/release-2024.04.30-28619/pkg/download-neco-deb/main.go#L32)
    * This should prepare the same `neco.deb` that is used in the staging data center.  We need to use the same algorithm with `neco-updater`.
    * I'm not sure whether we should update this just now or delay the update until we release this PR into stage0.  Probably **the former** because this command is used in the reproduced staging environment.
* [`find-installed-release`](https://github.com/cybozu-go/neco/blob/release-2024.04.30-28619/pkg/find-installed-release/main.go#L29)
    * This should return the same "datetime-rev" release name that is used in the staging data center.
    * I'm not sure whether we should update this just now or delay the update.  Probably **the latter** because this command is used to reproduce the staging environment.
    * [The caller](https://github.com/cybozu-go/neco/blob/release-2024.04.30-28619/pkg/find-installed-release/main.go#L29) should be updated too.

I may have overlooked something.

generator/generate_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-containerregistry/pkg/name"
+)
+
+func TestGetLatestImage(t *testing.T) {
+	t.Parallel()
+
+	type args struct {
+		repo            name.Repository
+		release         bool
```
morimoto-cybozu, line 21: This is not used because `getLatestImage()` no longer receives a `release` parameter.

generator/generate_test.go
```diff
@@ -0,0 +1,144 @@
...
+				release: true,
+			},
+			tagsList: []byte(`{"tags":["0.1.0","0.2.0", "1.0.0"]}`),
+			img: &neco.ContainerImage{
+				Name:       imageName(testRepository),
+				Repository: testRepository.String(),
+				Tag:        "1.0.0",
+				Private:    false,
+			},
+		},
```
morimoto-cybozu, line 96: These two cases are the same.

docs/artifacts.md
```diff
@@ -4,24 +4,22 @@ Artifacts
 [Artifacts](../artifacts.go) is the collection of components tested in [dctest](../dctest/), which consists of container images, debian package, and Flatcar image.
 
 There are two artifacts files.  When generating neco binaries and the neco deb package, `artifacts.go` is used
-by default.  `artifacts_release.go` is used only when `release` build tag is given.
+by default.  `artifacts_prod.go` is used only when `release` build tag is given.
 
 - [artifacts.go](../artifacts.go)
 
     This file describes artifacts to be tested in development.
```
morimoto-cybozu, line 11:
```suggestion
    This file describes artifacts to be tested in development and staging.
```

docs/artifacts.md
```diff
@@ -4,24 +4,22 @@ Artifacts
...
     It should be updated with `generate-artifacts` command.
     You may edit the file **ONLY IF** you would like to test different combination of components.
 
-- [artifacts_release.go](../artifacts.go)
+- [artifacts_prod.go](../artifacts.go)
 
-    This file is for CI.  The CI job updates this file by using
-    `generate-artifacts --release`.
+    This file describes artifacts to be tested in production.
```
morimoto-cybozu, line 17:
```suggestion
    This file describes artifacts to be used in production.
```

docs/artifacts.md
```diff
@@ -4,24 +4,22 @@ Artifacts
...
+    It should be updated with `generate-artifacts --release` command.
 
     The developers are PROHIBITED to modify this file.  This is to prevent
     merge conflicts in CI flow.
```
morimoto-cybozu, line 21: I don't think it is prohibited so strictly now.

PR #2570, opened 2024-06-04
Add artifacts_ignore.yaml temporarily

tkna:


zeroalphat (changes requested):
It seems that `artifacts_ignore.yaml` also includes container images that are not relevant.
For example, `ghcr.io/cybozu/cilium` is not used on `artifacts.go`.
Can you run the `generate-artifacts` command at hand and re-specify the container images that need to be ignored?

tkna:
recreated a PR
https://github.com/cybozu-go/neco/pull/2571

PR #2586, opened 2024-06-20
Fix broken path

takara9:


YZ775 (changes requested):
I found there are another broken path.
Could you fix these too?

https://github.com/cybozu-go/neco/blob/e068e9a88d853a3387df210ec5b1d0964d0c96b4/docs/dctest.md?plain=1#L64

https://github.com/cybozu-go/neco/blob/e068e9a88d853a3387df210ec5b1d0964d0c96b4/docs/dctest.md?plain=1#L71

YZ775 (changes requested):
@takara9 
I think changes of this PR is fine.
Could you squash commits?

YZ775 (approved):
LGTM

PR #2609, opened 2024-07-18
Update containerd version and modify the configuration

zeroalphat:
- Update containerd to 1.7.20.
- Enable image pull with sync fs in k8s-containerd config to prevent data inconsistencies in the event of an unexpected shutdown.
- Fix load-containerd-image script.

### DIFF
--- a/ignitions/common/files/opt/bin/load-containerd-image
+++ b/ignitions/common/files/opt/bin/load-containerd-image
@@ -9,7 +9,7 @@ img="$1"
 tag="$2"
 filename=/tmp/$(basename "$img")
 
-if /opt/bin/ctr --address=/var/run/k8s-containerd.sock --namespace k8s.io image inspect "$tag" >/dev/null 2>&1; then
+if /opt/bin/ctr --address=/var/run/k8s-containerd.sock --namespace k8s.io images list | grep "$tag" >/dev/null 2>&1; then
     exit 0
 fi
 
--- a/Makefile.common
+++ b/Makefile.common
@@ -2,7 +2,7 @@
 
 ## These should be updated when Kubernetes is updated
 CKE_VERSION = 1.28.6
-CONTAINERD_VERSION = 1.7.13
+CONTAINERD_VERSION = 1.7.20
 NERDCTL_VERSION = 1.7.6
 CRITOOLS_VERSION = 1.30.0
 RUNC_VERSION = 1.1.12
--- a/ignitions/common/files/etc/k8s-containerd/config.toml
+++ b/ignitions/common/files/etc/k8s-containerd/config.toml
@@ -50,6 +50,7 @@ oom_score = -999
   [plugins."io.containerd.grpc.v1.cri"]
     disable_tcp_service = true
     image_pull_progress_timeout = "0s"
+    image_pull_with_sync_fs = true
     stream_server_address = "127.0.0.1"
     stream_server_port = "10010"
     stream_idle_timeout = "4h0m0s"


ignitions/common/files/opt/bin/load-containerd-image
```diff
@@ -9,7 +9,7 @@ img="$1"
 tag="$2"
 filename=/tmp/$(basename "$img")
 
-if /opt/bin/ctr --address=/var/run/k8s-containerd.sock --namespace k8s.io image inspect "$tag" >/dev/null 2>&1; then
+if /opt/bin/ctr --address=/var/run/k8s-containerd.sock --namespace k8s.io images list "$tag" >/dev/null 2>&1; then
```
YZ775, line 12: I think `ctr image list` does not take image name as argument.
Instead, how about using `ctr image info "name==$tag"`?
but, it does not return exit 1 even if image does not exist. so we have to check length of output.


```console
 $ ctr image list -h
NAME:
   ctr images list - List images known to containerd

USAGE:
   ctr images list [command options] [flags] [<filter>, ...]

DESCRIPTION:
   list images registered with containerd

OPTIONS:
   --quiet, -q  Print only the image refs
 ```

below is result of command in dctst
```console
cybozu@rack1-cs5 ~ $ sudo ctr --address /var/run/k8s-containerd.sock -n k8s.io image list "name==hoge"
REF TYPE DIGEST SIZE PLATFORMS LABELS 
cybozu@rack1-cs5 ~ $ echo $?
0
cybozu@rack1-cs5 ~ $ 

cybozu@rack1-cs5 ~ $ sudo ctr --address /var/run/k8s-containerd.sock -n k8s.io image list "name==ghcr.io/cybozu-go/coil:2.7.0"
REF                          TYPE                                                 DIGEST                                                                  SIZE     PLATFORMS   LABELS                          
ghcr.io/cybozu-go/coil:2.7.0 application/vnd.docker.distribution.manifest.v2+json sha256:e1dd5464e242d1c816b022014abf0a81b1d8c0543c0a550fbc1614ad886b0ec8 96.3 MiB linux/amd64 io.cri-containerd.image=managed 
cybozu@rack1-cs5 ~ $ 

```

YZ775 (changes requested):

ignitions/common/files/opt/bin/load-containerd-image
zeroalphat: @YZ775 
Sorry, my confirmation was insufficient 🙇.
I thought it would be better to use `grep` as follows, because I thought it might be affected when the behavior of `ctr` is changed in the case of checking by length.

```
if /opt/bin/ctr --address=/var/run/k8s-containerd.sock --namespace k8s.io images list | grep "$tag"  >/dev/null 2>&1; then
    exit 0
fi
```
I think it works well enough when checking with length, but I thought it would be easier to read with grep.

YZ775: @zeroalphat I agree with the idea!

YZ775: I confirmed the idea works fine.

```
cybozu@rack1-cs5 ~ $ if sudo ctr --address /var/run/k8s-containerd.sock -n k8s.io image list | grep ghcr.io/cybozu-go/coil:2.7.0 > /dev/null 2>&1; then echo "image exsits"; fi
image exsits

cybozu@rack1-cs5 ~ $ if sudo ctr --address /var/run/k8s-containerd.sock -n k8s.io image list | grep invalid_image_name > /dev/null 2>&1; then echo "image exsits"; fi
cybozu@rack1-cs5 ~ $ 
```

zeroalphat: Thanks for the confirmation!
I have corrected it.

YZ775 (approved):
LGTM

PR #2614, opened 2024-07-22
Add cilium api-rate-limit config

chez-shanpu:
This PR adds cilium's api-rate-limit config.

note: Cilium doesn't support api-rate-limit helm value at 1.13 (it's supported from 1.15).


terassyi (approved):
LGTM

PR #2619, opened 2024-07-25
Set timeout to FW update commands

masa213f:
part of https://github.com/cybozu-private/neco-tasks/issues/397

This PR set the default timeaout and timeout option to `neco isoreboot` and `neco apply-firmware`.

Sometimes SSH to the worker machines hangs and the command stops.
To prevent this, I set a timeout to child command execution.

According to the execution log of past runs, execution time of `isoreboot`'s child command was about 12~13 seconds; `apply-firmware`'s child command was about 70 seconds, on worker nodes.
So I set the default timeout as follows:

- isoreboot: 30sec
- apply-firmware: 5min

### DIFF
--- a/pkg/neco/cmd/apply-firmware.go
+++ b/pkg/neco/cmd/apply-firmware.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/cybozu-go/log"
 	"github.com/cybozu-go/neco"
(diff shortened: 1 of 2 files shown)


morimoto-cybozu (approved):
I'm not sure whether `context.WithTimeout()` works well for `well.CommandContext()`.
It would be good to run the command manually for test with a very short timeout.
(This does not mean the need to update the CI test.)

PR #2635, opened 2024-08-07
Fix CI instability

yokaze:
This PR fixes three flaky behaviors to improve CI stability.

- test: retry neco bmc config set bmc-user
  Previously, `neco bmc config set bmc-user` sometimes fails because it fails to send packets and receives a message with `"missing HTTP content-type"`.
  This PR implements `execRetryAt` and `handleBMCRetry` to see if it is caused by connection error.

- neco-worker: wait changing systemd service state until transient state ends
  neco-worker was not deterministic at stopping systemd service, because `systemd` rejects `systemd stop <service_name>` with message `"Job for <service_name> canceled."` when it is already changing its state, especially when it is in `activating` or `deactivating` state.

- test: wait placemat
  In minor cases PID of `placemat2` is not observed and CI fails. This fix waits until `pgrep --exact placemat2` succeeds.


tkna (approved):
LGTM

PR #2637, opened 2024-08-08
Add feautre to perform daily reboot all nodes in dev environment

zeroalphat:
Add systemd service and timer for triggering daily reboot all nodes on dev environment.

### DIFF
--- a/debian/usr/bin/trigger-reboot-all-nodes
+++ b/debian/usr/bin/trigger-reboot-all-nodes
@@ -1,0 +1,23 @@
+#!/bin/sh
+NECO_COMMAND="/usr/bin/neco"
+CKECLI_COMMAND="/usr/bin/ckecli"
+YES_COMMAND="/usr/bin/yes"
+
+ENVIROMENT=$(${NECO_COMMAND} config get env)
+if [ "$ENVIROMENT" = "dev" ]; then
+    REBOOT_LIST=$(${NECO_COMMAND} rebooter list)
+    REBOOT_QUEUE=$(${CKECLI_COMMAND} rq list)
+   
+    # Confirm that the reboot list and reboot queue are empty
+    if [ "$REBOOT_LIST" = "null" ] && [ "$REBOOT_QUEUE" = "null" ]; then 
+        $NECO_COMMAND rebooter disable
+        sh -c $YES_COMMAND | $NECO_COMMAND rebooter reboot-worker --without-state retiring,retired
+        $NECO_COMMAND rebooter enable
+    else
+        echo "The reboot list or reboot queueu is not empty. Skipping reboot all nodes"
+        exit 0
+    fi
+else
+    echo "This script is only for dev environment"
+    exit 0
+fi
--- a/debian/usr/bin/trigger-reboot-all-nodes
+++ b/debian/usr/bin/trigger-reboot-all-nodes
@@ -1,23 +1,31 @@
 #!/bin/sh
-NECO_COMMAND="/usr/bin/neco"
 CKECLI_COMMAND="/usr/bin/ckecli"
+NECO_COMMAND="/usr/bin/neco"
+UNAME_COMMAND="/usr/bin/uname"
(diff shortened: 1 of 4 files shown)


debian/usr/bin/trigger-reboot-all-nodes
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+NECO_COMMAND="/usr/bin/neco"
+CKECLI_COMMAND="/usr/bin/ckecli"
+YES_COMMAND="/usr/bin/yes"
+
+ENVIROMENT=$(${NECO_COMMAND} config get env)
```
masa213f, line 6: Could you do early return?

```bash
ENVIROMENT=$(${NECO_COMMAND} config get env)
if [ "$ENVIROMENT" != "dev" ]; then
    echo "This script is only for dev environment"
    exit 0
fi

REBOOT_LIST=$(${NECO_COMMAND} rebooter list)
REBOOT_QUEUE=$(${CKECLI_COMMAND} rq list)
# Confirm that the reboot list and reboot queue are empty
if [ "$REBOOT_LIST" != "null" ] || [ "$REBOOT_QUEUE" != "null" ]; then 
    echo "The reboot list or reboot queueu is not empty. Skipping reboot all nodes"
    exit 0
fi

$NECO_COMMAND rebooter disable
sh -c $YES_COMMAND | $NECO_COMMAND rebooter reboot-worker --without-state retiring,retired
$NECO_COMMAND rebooter enable
```

debian/usr/bin/trigger-reboot-all-nodes
```diff
@@ -0,0 +1,23 @@
...
+if [ "$ENVIROMENT" = "dev" ]; then
+    REBOOT_LIST=$(${NECO_COMMAND} rebooter list)
+    REBOOT_QUEUE=$(${CKECLI_COMMAND} rq list)
+   
+    # Confirm that the reboot list and reboot queue are empty
+    if [ "$REBOOT_LIST" = "null" ] && [ "$REBOOT_QUEUE" = "null" ]; then 
+        $NECO_COMMAND rebooter disable
+        sh -c $YES_COMMAND | $NECO_COMMAND rebooter reboot-worker --without-state retiring,retired
+        $NECO_COMMAND rebooter enable
```
masa213f, line 15: Could you run these command on only one boot server?
I think we can use `neco rebooter leader` or `ckecli leader`.

masa213f (changes requested):
LGTM

debian/usr/bin/trigger-reboot-all-nodes
zeroalphat: I failed to take into consideration that it works on all boot servers.
I added a process to check if the server the script is running on is a leader.

debian/usr/bin/trigger-reboot-all-nodes
```diff
@@ -0,0 +1,31 @@
...
+LEADER=$(${NECO_COMMAND} rebooter leader)
+if [ "$HOSTNAME" != "$LEADER" ]; then
+    echo "This host is not a leader. Skipping reboot all nodes"
+    exit 0
+fi
+
+REBOOT_LIST=$(${NECO_COMMAND} rebooter list)
+REBOOT_QUEUE=$(${CKECLI_COMMAND} rq list)
+# Confirm that the reboot list and reboot queue are empty
+if [ "$REBOOT_LIST" != "null" ] && [ "$REBOOT_QUEUE" != "null" ]; then
```
masa213f, line 24: Is it correct? Should we change to `||`?

masa213f (changes requested):

debian/usr/bin/trigger-reboot-all-nodes
zeroalphat: Thank you for pointing this out.
I had overlooked the need to change the conditional judgement 🙇

masa213f (approved):
LGTM

PR #2640, opened 2024-08-13
test: retry idempotent neco bmc config set

yokaze:
This PR retries `neco bmc config set` operations when error message indicates a network error, because they are idempotent.


tkna (approved):
LGTM

PR #2654, opened 2024-09-03
fix neco-rebooter

YZ775:
fix neco-rebooter to 
- manage orphaned cke reboot queue entry
- hide processing group metrics when number of the reboot list entry is 0


masa213f (approved):
LGTM

PR #2663, opened 2024-09-06
fix reboot_test.go and collectEntries

YZ775:


dctest/reboot_test.go
```diff
@@ -229,6 +230,30 @@ func testNecoRebooterRebootGracefully() {
...
+		if err != nil {
+			Fail(fmt.Sprintf("failed to unmarshal json: %v", err))
+		}
+		racksTmp := []string{}
+		for _, node := range nodes.Items {
+			rack := node.Labels["topology.kubernetes.io/zone"]
+			if len(rack) == 0 {
+				continue
+			}
+			racksTmp = append(racksTmp, rack[len(rack)-1:])
```
masa213f, line 249: Why are you cut out numbers here?
In most reference points, `fmt.Sprintf("rack%s", racks[0])` is used. And it making the code difficult to read.

masa213f (changes requested):

dctest/reboot_test.go
YZ775: I decided to extract only rack numbers because the format of "rack" is different between application.

- k8s label: `topology.kubernetes.io/zone=rack1`
- sabakan option: `sabactl machines get --rack=1`

masa213f: I know. How about extracting the number only in neco-rebooter args?
Also, does `sh -c` work as expected?

```golang
rebootCmd := fmt.Sprintf("yes | neco rebooter reboot-worker --rack=%s", racks[0][4:] /* extract a rack number: "rackN" -> "N" */)
execSafeAt(bootServers[0], "sh", "-c", "'"+rebootCmd+"'")
```
or
```golang
execSafeAt(bootServers[0], "yes", "|", "neco", "rebooter", "reboot-worker", "--rack="+racks[0][4:] /* extract a rack number: "rackN" -> "N" */)
```

dctest/reboot_test.go
```diff
@@ -229,6 +230,26 @@ func testNecoRebooterRebootGracefully() {
...
+		}
+		slices.Sort(racksTmp)
+		racks := slices.Compact(racksTmp)
+		if len(racks) < 2 {
+			Fail("less than 2 racks are available")
+		}
```
masa213f, line 251: How about to use `Expect()` here too?

dctest/reboot_test.go
YZ775: I understand. and I fixed the code to extract number only in neco-rebooter command.

And, `sh-c` works fine.
It had been used before neco-rebooter was implemented like below.
https://github.com/cybozu-go/neco/blob/98bf903b8d9e278fd4c81767b4ed48895783810b/dctest/reboot_test.go#L83

dctest/reboot_test.go
```diff
@@ -317,18 +336,14 @@ func testNecoRebooterRebootGracefully() {
 		Expect(err).NotTo(HaveOccurred())
 
 		By("creating a pod")
-		stdout, stderr, err := execAtWithInput(bootServers[0], podYaml, "kubectl", "apply", "-f", "-")
-		if err != nil {
-			Fail(fmt.Sprintf("stdout: %s, stderr: %s, err: %v", stdout, stderr, err))
-		}
+		_, _, err = execAtWithInput(bootServers[0], podYaml, "kubectl", "apply", "-f", "-")
+		Expect(err).NotTo(HaveOccurred())
```
masa213f, line 340: Nice. But why don't you keep the `stdout` and `stderr`?
We can pass the optionalDescription, if we want.

https://github.com/onsi/gomega/blob/7cabed651e981da0c7d0219ad7208626cef58016/gomega_dsl.go#L524

dctest/reboot_test.go
```diff
@@ -346,16 +361,16 @@ func testNecoRebooterRebootGracefully() {
...
 
-		By("waiting for skipping rack1 and moving to rack2")
+		By(fmt.Sprintf("waiting for skipping %s and moving to %s", racks[0], racks[1]))
 		Eventually(func() error {
 			stdout, stderr, err := execAt(bootServers[0], "neco", "rebooter", "show-processing-group")
 			if err != nil {
 				return fmt.Errorf("stdout: %s, stderr: %s, err: %v", stdout, stderr, err)
 			}
-			if string(stdout) != "rack2\n" {
+			if strings.Contains(string(stdout), racks[1]) {
```
masa213f, line 373: How about use `strings.TrimSpace`?

masa213f (changes requested):

masa213f (approved):
LGTM

PR #2671, opened 2024-09-10
dctest: Fixed wrong use of StopTrying

masa213f:
This PR fixed wrong use of `StopTrying()`.
In this case, we may need to call `StopTrying().Now()`.

https://pkg.go.dev/github.com/onsi/gomega#pkg-variables


yokaze (approved):
LGTM
Thank you for the correction 🙇🙇

PR #2672, opened 2024-09-10
Securiy fix cloudflare/circ

takara9:
Apply following security fix

- https://github.com/cybozu-go/neco/security/dependabot
  - https://github.com/cybozu-go/neco/security/dependabot/11
  - https://github.com/cybozu-go/neco/security/dependabot/19


masa213f (approved):
LGTM

PR #2688, opened 2024-10-08
increase BOOT_DISK_SIZE

YZ775:
CI sometimes fails due to rack of free disk.
This PR increases BOOT_DISK_SIZE to 30GB from 20GB.

After this PR will be merged, we need to merge following PR.
- https://github.com/cybozu-go/neco-gcp/pull/153


masa213f (approved):
LGTM

PR #2698, opened 2024-10-15
Allow disabling daily reboot

masa213f:
This PR allows users to turn off the daily reboot.
If users create `/tmp/no-reboot` on all boot servers, the daily reboot will be disabled.


zeroalphat (approved):
LGTM

PR #2724, opened 2024-11-27
Disable endpoint-manager's gc in cilium-agent

chez-shanpu:
"endpoint-manager" which is a subsystem in cilium-agents has GC system.
This GC sometimes removes healthy CEP.
(I think this is due to poor coordination with Coil, but the root cause is not currently known.)

It was implemented as a way to deal with the bug with dockershim.
However, dockershim is not used in Neco and as far as I looked up I cannot find the CEP that has to be garbage collected.

This PR disables endpoint-manager's GC and avoid deleting healthy CEP.

### DIFF
--- a/cilium/pre/values.yaml
+++ b/cilium/pre/values.yaml
@@ -20,6 +20,8 @@ enableIPv4Masquerade: false
 enableIdentityMark: false
 externalIPs:
   enabled: true
+extraArgs:
+  - "--endpoint-gc-interval=0"
 extraConfig:
   bpf-ct-timeout-regular-any: 1h0m0s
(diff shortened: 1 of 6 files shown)


cilium/pre/values.yaml
```diff
@@ -20,6 +20,9 @@ enableIPv4Masquerade: false
 enableIdentityMark: false
 externalIPs:
   enabled: true
+extraArgs:
+  - "--endpoint-gc-interval"
+  - "0"
```
yokaze, line 25: I suspect it would be easier to read?
```suggestion
  - "--endpoint-gc-interval=0"
```

tkna (approved):
LGTM

PR #2736, opened 2024-12-04
Fix cpu-governor.service to set performance governor for existing CPU

YZ775:
If we set SysProfile to `PerfOptimized`, there are no files in `/sys/devices/system/cpu/cpu*/cpufreq/scaling_governor`.
In such situation, we want to skip cpu-governor.service.


masa213f (approved):
LGTM

PR #2752, opened 2024-12-18
Use token to download a debian package

zoetrope:
- Set bearer token in header.
- Change download url from `browser_download_url` to `url`. Because `browser_download_url` cannot be used with token.

ref. https://github.com/cybozu-private/neco-tasks/issues/529


masa213f (approved):
LGTM

PR #2755, opened 2024-12-20
Add boot network interface

masa213f:
This PR add `boot` network interface to boot servers.

### DIFF
--- a/installer/setup/network.go
+++ b/installer/setup/network.go
@@ -7,7 +7,9 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"slices"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/cybozu-go/neco"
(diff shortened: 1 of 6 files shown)


installer/setup/network.go
```diff
@@ -76,13 +93,21 @@ Kind=dummy
 		return fmt.Errorf("failed to create netdev file for %s: %w", name, err)
 	}
 
-	s = fmt.Sprintf(`[Match]
+	if len(addrs) == 0 {
```
masa213f, line 96: In the case of the `boot` interface, I only want to create the .netdev file.

installer/setup/network.go
```diff
@@ -76,13 +93,21 @@ Kind=dummy
...
 
 [Network]
-Address=%s/32
-`, name, addr.String())
-	err = os.WriteFile(fmt.Sprintf("/etc/systemd/network/%s.network", name), []byte(s), 0644)
+`, name)
+	builder.WriteString(head)
+	for _, a := range addrs {
+		builder.WriteString(fmt.Sprintf("Address=%s/32\n", a.String()))
+	}
```
masa213f, line 109: If multiple addresses are passed, this builder will output the following files.
`Address=` will be specified more than once. 
```
[Match]
Name=test

[Network]
Address=10.20.30.40/32
Address=10.20.30.41/32
```

This may work fine.

https://manpages.debian.org/bookworm/systemd/systemd.network.5.en.html#%5BNETWORK%5D_SECTION_OPTIONS

> Address=
> A static IPv4 or IPv6 address and its prefix length, separated by a "/" character. **Specify this key more than once to configure several addresses.** The format of the address must be as described in [inet_pton(3)](https://manpages.debian.org/bookworm/manpages-dev/inet_pton.3.en.html). This is a short-hand for an [Address] section only containing an Address key (see below). This option may be specified more than once.

zeroalphat (approved):
LGTM

PR #2776, opened 2025-01-16
implement automatic night release feature

YZ775:
I'll not merge this PR until k8s 1.30 is applied.

### DIFF
--- a/pkg/neco/cmd/config_get.go
+++ b/pkg/neco/cmd/config_get.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/cybozu-go/log"
 	"github.com/cybozu-go/neco"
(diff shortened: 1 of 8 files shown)


pkg/neco/cmd/config_get.go
```diff
@@ -125,6 +128,21 @@ Possible keys are:
 					return err
 				}
 				fmt.Println(ipBlock)
+			case "release-time":
+				t, err := st.GetReleaseTime(ctx)
+				if err != nil {
+					return err
+				}
+				ss := strings.Split(t, ",")
```
morimoto-cybozu, line 136: A single cron schedule can contain commas, so another type of spliter could be appropriate.
I think it's OK to add this behavior to limitations.

YZ775: Thank you. I fixed it to use `\n` as a separator.

morimoto-cybozu (approved):

PR #2778, opened 2025-01-16
test: show additional information when CKE fails to complete

yokaze:
This PR prints additional information when CKE fails to complete (especially in case of updating Cilium).


chez-shanpu (approved):
LGTM

PR #2788, opened 2025-01-29
sabakan-state-setter: detect machine failures using Prometheus alerts

morimoto-cybozu:
See-also: https://github.com/cybozu-private/neco-tasks/issues/466


pkg/sabakan-state-setter/controller.go
```diff
@@ -271,9 +288,12 @@ func (c *Controller) runOnce(ctx context.Context) error {
 	for _, m := range machines {
 		newState, ok := newStateMap[m.Serial]
 		switch {
-		case !ok || newState == m.State:
+		case !ok || newState == m.State || (newState == stateUnhealthyImmediate && m.State == sabakan.StateUnhealthy):
 			c.ClearUnhealthy(m)
 			continue
+		case newState == stateUnhealthyImmediate:
+			c.ClearUnhealthy(m)
+			newState = sabakan.StateUnhealthy
```
zeroalphat, line 296: Could you please explain why are you using the `stateUnhealthyImmediate` variable and executing `ClearUnhealthy`? I'm not quite clear on their purpose.

pkg/sabakan-state-setter/controller_test.go
```diff
@@ -34,6 +36,22 @@ var machineTypeSerfOnly = &machineType{
...
+	{
+		Name:        "DiskNotRecognized",
+		SerialLabel: "serial",
+		State:       sabakan.StateUnhealthy,
+		Labels: map[string]string{
+			"severity": "error",
+		},
+	},
+	{
+		Name:         "LLPDDown",
```
zeroalphat, line 49: Is this LLDPDown?

pkg/sabakan-state-setter/controller_test.go
```diff
@@ -148,12 +196,36 @@ hw_storage_controller_status_health{controller="PCIeSSD.Slot.2-C", system="Syste
...
+				},
+			},
 			expected: map[string]sabakan.MachineState{
 				"00000000": sabakan.StateUnhealthy,   // one of two CPUs is issuing a warning
 				"00000001": sabakan.StateUnhealthy,   // all HDD are unhealthy; # of healthy HDDs falls below MinimumHealthyCount (0 < 1)
-				"00000002": sabakan.StateHealthy,     // one of two HDDs is unhealthy, but it is acceptable
+				"00000002": sabakan.StateHealthy,     // one of two HDDs is unhealthy, but it is acceptable; alert "DiskNotRecognized" is firing but it is still warning
 				"00000003": sabakan.StateUnreachable, // serf status is "failed"
+				"00000004": sabakan.StateUnhealthy,   // alert "DiskNotRecognized" is firing; grace period is ignored
+				"00000005": sabakan.StateUnreachable, // alert "LLPDDown" is firing; alert "DiskNotRecognized" is ignored because it is less severe
```
zeroalphat, line 228: Ditto

pkg/sabakan-state-setter/controller.go
morimoto-cybozu: @zeroalphat 
It's because I want to skip the [grace period](https://github.com/cybozu-go/neco/blob/release-2025.02.04-34486/docs/sabakan-state-setter.md#grace-period-of-setting-unhealthy-state).

Currently `sabakan-state-setter` changes a machine's state to `unhealthy` only after waiting for the grace period.
This seems reasonable when the cause of the change is `serf` status or `monitor-hw` metrics.
In this pull request, I add state changes caused by `Prometheus` alerts.
In my design, the grace period is not necessary for this new state change because we can configure Prometheus alerts to wait a sufficient amount of time before becoming active.
https://github.com/cybozu-go/neco/pull/2788/files#diff-819895dee5e92a2356096850816be02af8e2b41ea1f6f69832d71967303465f3
So I added a new state candidate `stateUnhealthyImmediate` for the new state change.

The controller manages the start timestamps of to-be-unhealthy machines.
If a machine is newly evaluated to `StateUnhealthy` (not `stateUnhealthyImmediate`), the controller registers the timestamp via `Controller.RegisterUnhealthy()`.
The controller continuously checks the timestamp while the machine is evaluated to `StateUnhealthy` and the current status is not `StateUnhealthy`.
Once the controller has changed the status to `StateUnhealthy`, it deletes the timestamp via `Controller.ClearUnhealthy()`.
The controller also deletes the timestamp when it becomes unnecessary.
If the new state candidate is `stateUnhealthyImmediate`, we don't need the timestamp any more.
So I add `c.ClearUnhealthy(m)`.
We can call `Controller.ClearUnhealthy()` without checking that a timestamp is registered actually.

morimoto-cybozu:
I've rebased to reflect new `artifacts.go`.

pkg/sabakan-state-setter/controller.go
zeroalphat: Thank you for explaining that. 
I now have a clear understanding of why the `stateUnhealthyImmediate` variable needs to be added.

zeroalphat (approved):
LGTM

PR #2791, opened 2025-01-30
fix systemd services

YZ775:
fix systemd services to restart infinitely.
upgrade-test definitely fails now, so I'll edit branch protection rule and merge this PR after I get approve.


zeroalphat (approved):
LGTM

PR #2792, opened 2025-01-30
Remove unnecessary neco-rebooter from worker

YZ775:
neco-rebooter is restarted by postinst in .deb
So restarting by worker is unnecessary.


masa213f (approved):
LGTM

PR #2793, opened 2025-02-03
Update flatcar to 4152.2.0

zeroalphat:


YZ775:
It's described in release note. So I think it's better to fix some ignition templates that uses binary under `/usr/sbin`
```
The /usr/sbin directory is now merged into /usr/bin, so the former became a symlink
```

https://github.com/search?q=repo%3Acybozu-go%2Fneco+path%3A%2F%5Eignitions%5C%2F%2F+%2Fusr%2Fsbin&type=code

YZ775 (approved):
LGTM

PR #2796, opened 2025-02-13
Fix no-output termination

yokaze:
CircleCI terminates test programs when they run without output longer than a certain limit (default: 10m).
On the other hand, some Ginkgo jobs should wait for some time to see completion of slow work.

Therefore, we need to tell CircleCI to avoid terminating jobs longer than the period set to Ginkgo jobs.
This PR clarifies the division of responsibility:
- Ginkgo jobs can use round number
- CircleCI should wait for an extra time (set to 2m)


tkna (approved):
LGTM

PR #2799, opened 2025-02-19
Split the test for artifact.go from main to check-artifacts

chez-shanpu:
This PR splits the test for artifact.go from main and makes it not required.


yokaze (approved):
LGTM

